PR #3, opened 2025-03-08
[FEATURE] De-fortranify

jsherer:
From: https://github.com/bazineta/js8call


jsherer:
Merge to include defortranify and qt6 updates

PR #24, opened 2025-07-05
Prevent Heartbeat responses when m_messageBuffer is not empty

Chris-AC9KH:
This is a bug fix that prevents unwanted auto Heartbeat responses when there is incoming text in the message pane and a callsign is not selected. Most experienced users who do a lot of messaging and relays typically uncheck 'Enable Heartbeat Acknowledgements (ACK)' to prevent broken incoming. They can now leave that checked, if desired.

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -7709,7 +7709,7 @@ void MainWindow::updateModeButtonText(){
     auto multi = ui->actionModeMultiDecoder->isChecked();
     auto autoreply = ui->actionModeAutoreply->isChecked();
     auto heartbeat = ui->actionModeJS8HB->isEnabled() && ui->actionModeJS8HB->isChecked();
-    auto ack = autoreply && ui->actionHeartbeatAcknowledgements->isChecked() && (!m_config.heartbeat_qso_pause() || selectedCallsign.isEmpty());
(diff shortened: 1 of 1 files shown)


Chris-AC9KH:
This PR has been tested and I recommend for merge to master

Chris-AC9KH:
This PR has been rebased on main since a subsequent PR was merged first.

jsherer (approved):
Simple fix, great work.

Chris-AC9KH:
@jsherer there's been a further request to disable auto-replies to other queries, especially in group situations where an incoming may be broken by a SNR?, HEARING? etc query and EVERYBODY replies to it. I have a fix for that and it's being tested. But using the state of the m_messageBuffer is extremely effective in this regard as it resets instantly when the EOT character is generated, and waits for a bit of time if we get a dropped frame.

I am NOT queueing auto-replies to these queries because in my experience it can take 5 minutes or more for a MSG to be received, even more time if Slow mode is being used in weak signal situations. It's a moot point by then and too far out of the time frame for a queued reply to be useful. After suitable testing of the latest changes to this, if I'm satisfied with it I'll submit another PR for that.

jsherer:
Sounds good here. Give'em what they want :D

Joe-K0OG:
Most Excellent - thanks Chris @Chris-AC9KH for sticking with it!

73,
-Joe-
K0OG

Issue #5, opened 2025-03-08
[DOC] Improve documentation for building


jsherer:
## Prerequisite Install Instructions
- [ ] OSX
- [ ] Linux
- [ ] Windows

## Build Instructions
- [ ] OSX
- [ ] Linux
- [ ] Windows

MidwestAire:
@jsherer I submitted a PR that is a start on this. It is not intended to be merged right away.

Background info:
There was too many README's in the sources. So I moved the summary of changes from 2.2 -> 2.3, written by @bazineta , to the default README.md that displays on the JS8Call repo. Then linked that to three new BUILD-xxx markdown files for MacOS, Linux and Windows. To see how this looks when displayed on GitHub you can view these two links:

The default README.md always displays at the bottom of the source code repo
https://github.com/MidwestAire/js8call/tree/pr/build_documentation

This is the MacOS one I just cleaned up and did for build documentation
https://github.com/MidwestAire/js8call/blob/pr/build_documentation/BUILD-MacOS.md

@Joe-K0OG would have to look at this and supply the information for building on Linux and Windows. I could "wing it" and try to do it. But I'm a MacOS guy and while I could probably do the Linux page ok, Windows would be a disaster since I don't even know how to start up a Windows box.

So this is, again, just a start. Other developers that want to jump in, feel free to add commits to the PR for the rest of it.

Chris AC9KH

Joe-K0OG:
I added some first-draft of instructions for building on Windows and Linux, mostly copy/paste from an earlier post ( see https://github.com/js8call/js8call/issues/4#issuecomment-2709047799).  Someone else jump in and clean it up if needed!  I'm not very familiar with MD (much more so with HTML, so resorted to some HTML to make it work better).

73,
-Joe-
K0OG

Joe-K0OG:
Chris,

Those are from notes I made for Jordan some time ago, so I had something to start with.  I think they are fairly accurate, but am assuming that if one will build from source, they have some basic grasp of operating in command-line windows and the basics of the operating system environments, so not all the minute details are there.  I think it is good to go.

Thanks for all your help!

73,
-Joe-
K0OG

MidwestAire:
@jsherer @Joe-K0OG I think this PR (https://github.com/js8call/js8call/pull/35) is finalized for merging with the main branch. I squashed it down into three commits, so that should make a cleaner commit history.

Hopefully, this issued can be closed.

Chris AC9KH

[...]

PR #35, opened 2025-07-19
Docs/Improve Build Documentation

Chris-AC9KH:
This is an initial commit and PR to improve the build documentation for MacOS, Linux and Windows. It adds a new docs directory and includes a new Contributors Guide. The old INSTALL and INSTALL-WSJTX files are obsolete and removed.

Addresses issue https://github.com/js8call/js8call/issues/5


MidwestAire:
Went thru the MacOS build instructions and tested the copy & paste commands as it would be displayed on the webpage from start to finish. Corrected some typo's and it all works. So I think this PR is ready to go.

Chris-AC9KH:
I agree. Both INSTALL and INSTALL-WSJTX are legacy files that could probably be deprecated The INSTALL file wasn't even complete for Windows and Mac in the first place. @Joe-K0OG what do you think, Joe? If you think they should be deprecated, I'll submit that as a separate commit to the PR, and if Jordan wants to keep them for odd reason, he can?

Chris AC9KH

Joe-K0OG:
Chris,

I agree that the content of the old INSTALL files are not of any value at this point, since that has all been replaced by our PR.  If someone wants to see the old files, they can look at the old source.

73,
-Joe-
K0OG

BUILD-Linux.md
```diff
@@ -0,0 +1,44 @@
...
+  sudo apt build-dep js8call
+  
+Fedora:
+  sudo dnf groupinstall "Development Tools" "Development Libraries"
+  sudo dnf build-dep hamlib
+  sudo dnf build-dep js8call</pre>
+
+**Building Hamlib**
+
+It is easiest to build Hamlib 4.6.4 and install it on the system.  To see what versions of Hamlib are available, while in the hamlib directory, type "git branch -r" or "git tag" to get a list.  To get the source and build Hamlib, open a command prompt and issue the following commands:
```
aknrdureegaesr, line 21: I could build JS8Call nicely with the Hamlib that can be installed via `apt install libhamlib-dev` on Debian Bookworm, Debian version `4.5.4-1+b1`. The resulting self-compiled JS8Call seems to work fine for me. So I think we should either outright remove this section or else mention under what circumstances this procedure is of an advantage, compared to simply using what's in the box.

aknrdureegaesr: This topic can be dropped here, it now treated by an issue of itself over at #41.

BUILD-Linux.md
```diff
@@ -0,0 +1,44 @@
+------------------------------------------------------------------------------
+# Building JS8Call on Linux
+
+
+**Prerequisites**
```
aknrdureegaesr, line 5: This is a headline and could be marked as

    ## Prerequisites

BUILD-Linux.md
```diff
@@ -0,0 +1,44 @@
...
+
+<u>Linux Versions:</u>  Debian 12 or newer (or derivatives, e.g., Mint 22.1, Ubuntu 24.04, MX Linux); Fedora 39 or newer.  Enable Source Code archives in the repositories, then issue the following commands at the command-line prompt:
```
aknrdureegaesr, line 7: Here, `**Linux Versions:**` would be better Markdown.

BUILD-Linux.md
```diff
@@ -0,0 +1,44 @@
...
+
+<pre>Debian & derivatives:
+  sudo apt install build-essential
+  sudo apt build-dep hamlib
+  sudo apt build-dep js8call
```
wmiler, line 12: Not sure why these are here, and it won't work unless you have updated the /etc/apt/sources.list.d/ files to include a deb-src in the Types: block for each package. 
Recommend you either remove the build-dep's or go into more detail on how to setup it up.

BUILD-Linux.md
```diff
@@ -0,0 +1,44 @@
+------------------------------------------------------------------------------
+# Building JS8Call on Linux
+
+
+**Prerequisites**
+
+<u>Linux Versions:</u>  Debian 12 or newer (or derivatives, e.g., Mint 22.1, Ubuntu 24.04, MX Linux); Fedora 39 or newer.  Enable Source Code archives in the repositories, then issue the following commands at the command-line prompt:
+
+<pre>Debian & derivatives:
+  sudo apt install build-essential
```
wmiler, line 10: A little short on everything that needs to be installed.
``` shell
sudo apt install -y build-essential git cmake libhamlib-dev qt6-base-dev qt6-multimedia-dev qt6-serialport-dev libusb-1.0-0-dev libfftw3-dev libboost-all-dev
```

BUILD-Linux.md
aknrdureegaesr: This is an old trick that doesn't presently work for `js8call`.

The trick works if a distribution has foo 4.5.6 and you want to compile 4.5.7. Then, the `apt build-dep foo` will pull in all the development versions of libraries that were required to build 4.5.6, and you can hope that that's the same that will be needed to compile 4.5.7 as well.

In our case, it is likely to pull in the build dependencies of jscall 2.2.x, so, wrong versions of stuff, e.g., QT 5 instead of QT 6.

So indeed, this is not helpful and should be removed. I'm with you as far as that is concerned.

Chris-AC9KH:
@Joe-K0OG I converted the Windows and Linux files to markdown. I think I got all the HTML tags removed so they display correctly. You can check them here:
https://github.com/Chris-AC9KH/js8call/tree/pr/build_documentation/docs

@aknrdureegaesr @wmiler @dl1jbe could you please review these and suggest any changes before I change this from WIP to PR to commit. Thanks!

Joe-K0OG:
@Chris-AC9KH Good work Chris - they look good to me!

(Markdown is new to me - I'm an old HTML guy accustomed to using vi from the early days of HTML in the late '80s... ;) )

73,
-Joe-
K0OG

BUILD-Linux.md
Chris-AC9KH: It looks to me like sudo apt build-dep js8call should probably be replaced with: sudo apt install -y build-essential git cmake libhamlib-dev qt6-base-dev qt6-multimedia-dev qt6-serialport-dev libusb-1.0-0-dev libfftw3-dev libboost-all-dev libudev-dev libxkbcommon-dev

Are we in agreement on that? When I tried Joe's instructions they worked on arm64 install of Ubuntu 25. But on Ubuntu 24 it probably won't work without listing out the full dependency list.

Chris-AC9KH: I changed the command for Debian. I hope it looks right. I'm not sure about Fedora. I've never run Fedora in my life.

Joe-K0OG: I'll test this on some pristine systems tomorrow to make sure it works.  I think it will, but need to test to be sure.

73,
-Joe-
K0OG

Chris-AC9KH: I know it worked on arm64 Ubuntu 25. But I don't know about Ubuntu 24 or any other system that would try to pull deps for JS8Call 2.2.0. I changed the command to what @wmiler suggested, as I think that one will work no matter what. Maybe on a pristine install try it as it's shown here to make sure that works.

https://github.com/Chris-AC9KH/js8call/blob/pr/build_documentation/docs/BUILD-Linux.md

lordmorgul:
For Fedora 42, with dnf5 the "groupinstall" is not recognized and should be "group install", with an added space.
The groups referenced are no longer found, the naming changed with dnf5 as well.
The groups should be "development-tools" and "c-development". The quotes are not needed as all group names now have no whitespace.

`sudo dnf group install development-tools c-development`

The following "dnf build-dep" commands work fine on F42.

I'm not sure how far back Fedora moved to dnf5 vs dnf4 (maybe F38? or F39?, so been awhile). It would seem reasonable to just include the latest for dnf5.
https://dnf5.readthedocs.io/en/latest/commands/group.8.html

docs/BUILD-Linux.md
```diff
@@ -0,0 +1,40 @@
...
+   `sudo ldconfig` (this is necessary to load/configure system libraries)\
+   `rigctl --version` (this is to test that Hamlib is installed and at the correct version level)
+
+**Building JS8Call**\
+To obtain the JS8Call source and build, issue the following commands in a terminal window:\
+   `cd Downloads`\
+   `git clone https://github.com/js8call/js8call.git js8call`\
+   `cd js8call`\
+   `mkdir build && cd build`\
+   `make ..`\
```
lordmorgul: Missing cmake here, otherwise get "make: Nothing to be done for '..'."

add `cmake ..`

Correction: it was just typo, `make ..` was supposed to be `cmake ..`

Then `make -j 4` works within build

lordmorgul: then the `make ..` needs to be `make` since we are in build at this point

docs/BUILD-Linux.md
```diff
@@ -4,18 +4,23 @@
...
+   sudo apt install -y build-essential git cmake libhamlib-dev qt6-base-dev \ 
+                       qt6-multimedia-dev qt6-serialport-dev libusb-1.0-0-dev \
+                       libfftw3-dev libboost-all-dev libudev-dev libxkbcommon-dev
+```
+```
+   sudo apt build-dep hamlib
+```
   
 Fedora:\
    `sudo dnf groupinstall "Development Tools" "Development Libraries"`\
```
lordmorgul: `sudo dnf group install development-tools c-development`

docs/BUILD-Linux.md
Chris-AC9KH: Ok, thanks. I verified that this works on arm64 Ubuntu 25. Since that install is running in the UTM virtual machine on an Apple silicon Mac, and I'm not sure that arm64 linux totally understands Apple's core architecture, I left the `-j 4` off the make command. It only took it 27 seconds to build anyway.

docs/BUILD-Linux.md
Chris-AC9KH: I changed that line but don't have an install of Fedora to test it on.

Chris-AC9KH:
I force pushed the branch with a second commit to fix the linux build commands.

https://github.com/js8call/js8call/pull/35/commits/beda59168cea9fecbb72b9fef8d7ed92459d03bb

Tested the Debian portion on arm64 Ubuntu 25. Anybody that can test the Fedora portion and verify it?

Before I change this from WIP I'll probably reformat the linux page, combine multiple commands with && and make the commands copy-able with the markdown without using highlight and copy.

wmiler:
@Joe-K0OG Question: I'm installing QT via msys2 pacboy and dumpcpp is named dumpcpp-qt6, do you see that on your Windows build? I don't see it being reassigned in JTSDK.

Joe-K0OG:
@wmiler Wyatt, I don't remember "dumpcpp" but I install Qt using the Qt MaintenanceTool.exe, then point JTSDK64-Tools to that C:\Qt\Qt6.6.3 install directory.  Looking at C:\Qt\6.6.3\mingw_64\bin, there is a dumpcpp.exe that has been installed.  

When installing, many components are installed as they scroll by, and I don't watch to see what is installed.  Boring...  Ha!  ;)

73,
-Joe-
K0OG

Joe-K0OG:
I've spent some time today testing the Linux build instructions on pristine installations, and find that several changes need to be made for it to be complete, and work properly.  I was going to edit the document, but don't have write access.  Here are my notes:

## Debian & derivatives:
* Add a line "sudo apt update" before the "sudo apt install -y ..." -- Necessary to update the repository database
* Replace the "`libboost-all-dev`" with "`libboost*1.81*-dev`" -- Without specifying version 1.81, it will pull in 1.74 which doesn't work.

## Fedora:
* Add a line "sudo dnf update" before the other lines -- Necessary to update the repository database
* The "sudo dnf" commands were good as they were for DNF version 4, but DNF version 5 requires a new format.  I've also added some components which are needed (where I don't specify DNF version, it works for both versions):
 ```
  sudo dnf groupinstall "Development Tools" "Development Libraries" libqt6* qt6* rpmbuild -- for RPM v4
  sudo dnf install @development-tools @development-libraries libqt6* qt6* rpmbuild -- for RPM v5
  sudo dnf build-dep hamlib
  sudo dnf build-dep js8call
  sudo dnf remove hamlib -- to remove old hamlib components which were installed by the previous build-dep
```
(Note:  The above "libqt6*" and "qt6*" DNF installs pull in Qt6.8.3 as of today.  If a different version of Qt is desired or needed, other measures will need to be taken, e.g., manually installing Qt using the Linux Qt installer, or explicitly pull a different version from the Fedora repository if available.)

## Building Hamlib:
* Change "make -j 4" to "cmake --build . -- -j 4" -- This works better on both Debian and Fedora, especially for packaging

## Building JS8Call:
* Change "make .." to "cmake .." -- this was a typographical error on my part
* Add a line at the end (after the "make -j 4"), "sudo make install" -- this will install the resulting build on the local system

## Packaging JS8Call:
* Perhaps a separate section should be added concerning packaging.
* In my testing on Fedora 39, all works as specified ("cpack -G RPM"), but was buggy on Fefora 41.  This is a work-in-progress.  What I find does work on Fedora is the following:
  .Build hamlib in ~Downloads/hamlib with the following script (assuming the source has been obtained with git, and "git checkout" has been used to select the hamlib version, per above, but placed is ~Downloads/hamlib/src):
```
cd ~Downloads/hamlib/src
./bootstrap
mkdir build && cd build
../src/configure --prefix=~Downloads/hamlib --disable-shared --enable-static --without-cxx-binding --disable-winradio CFLAGS="-g -O2 -fdata-sections -ffunction-sections" LDFLAGS="-Wl,--gc-sections"
make -j 4
make install-strip
cd bin && ./rigctl --version
exit
```
Then to build JS8Call, I use the following modified cmake (instead of the "cmake .." in the JS8Call build instructions) to point to the specially-built hamlib above:
`cmake -D CMAKE_PREFIX_PATH=~Downloads/hamlib -D CMAKE_INSTALL_PREFIX=/opt/js8call ..`

I actually use the same scripts similar to the above on both Debian derivatives and Fedora, and they work flawlessly, and make the build process faster for me.  Perhaps someone else can streamline this further.
     
* Should we add a section on building an AppImage?  I tried Docker on Debian, but had problems.  I have not spent much time looking into this, but have been successful building AppImages on Debian derivatives with Podman.  By editing some of the files in the JS8Call docker directory, the Hamlib version can be controlled, and by using "git checkout" on the JS8Call source directory, the JS8Call version can be controlled.  Thus, a stable-release version of JS8Call including the stable-release of Hamlib can be generated.

73,
-Joe-
K0OG

**Comment by Chris-AC9KH** [comment]
Thanks for that Joe. I'll make the necessary corrections to the Linux file.

I don't know that .AppImage really applies here. Probably not even .deb or .rpm because I don't think that people reading it and building the code are doing it for distribution of a binary. More likely to try to compile and run on their local system because they might have a different architecture like arm64 on a Raspberry Pi that the production builds aren't built for.

That's what I think, I don't know for sure. But if my hunch in correct then it wouldn't be worth it to add .AppImage build instructions. It also wouldn't hurt to add them, so if anybody wants to add that we will.

**Comment by Joe-K0OG** [comment]
>I don't know that .AppImage really applies here. Probably not even .deb or .rpm because I don't think that people reading it and building the code are doing it for distribution of a binary.

That makes some sense.  Maybe another document referenced by the contributor's guidelines that discusses packaging would be reasonable.

I just realized that the command to install on the local system is missing.  The following line should be added at the end of the "Building JS8Call" instructions:

`sudo make install`

73,
-Joe-
K0OG

**Comment by Joe-K0OG** [comment]
>I don't know that .AppImage really applies here. Probably not even .deb or .rpm because I don't think that people reading it and building the code are doing it for distribution of a binary. More likely to try to compile and run on their local system because they might have a different architecture like arm64 on a Raspberry Pi that the production builds aren't built for.

That makes some sense @Chris-AC9KH Chris.  Maybe a separate document on packaging can be referenced in the contributor's guidelines.

73,
-Joe-
K0OG

**Comment by Chris-AC9KH** [comment]
@Joe-K0OG so, none of these commands are working. I found a desktop ISO for Ubuntu 24.04.3LTS arm64 and the very first command fails for Debian systems. For instance, in Ubuntu the package needed for boost is actually libboost1.83-dev

Screenshot 2025-08-19 at 20 51 22

It might be different in Debian, but very few people actually run Debian. Ubuntu has the lion's share of the market in Debian-based systems.

The second command fails because of no src URL's in sources.list

Screenshot 2025-08-19 at 21 08 47

While this is easy for me fix, the average person who wants to build JS8Call for their arm64 Raspberry Pi is going to have problems without more complete instructions that cover every single step like I did for MacOS.

**Comment by Chris-AC9KH** [comment]
@Joe-K0OG JS8Call runs on Ubuntu 24 arm64 (also verified on Ubuntu 25 arm64). But I'm going to start over with a fresh system again and document what I had to do to get it to build. Ultimately, going to Qt Creator is going to be better. There's already a js8call.pro file in the source tree, which is a project file for Qt Creator but it hasn't been updated since like 2011 and it's for the old Fortran build from WSJTx

Screenshot 2025-08-19 at 22 25 11

Actually, JS8Call does not need boost to run - only to build it. libusb-1.0.0 got installed probably with libhamlib-dev. But we're missing the scalable vector graphics framework from Qt and it's no wonder the UI looks like crap compared to MacOS.

So I have to do a little debugging with Qt Creator and it may be necessary to pull Qt6 from github, or install it with Qt Creator to get a proper build instead of using distro-packaged Qt libraries. It's just not consistent from one distro of linux to another and this has been a problem in linux since the word "go". There's a reason I stopped using linux for anything useful 10 years ago :-)

**Comment by Joe-K0OG** [comment]
Chris @Chris-AC9KH,
The Boost problem is due to me not reading my own handwriting clearly.  What I meant to put there was `"libboost*1.8*-dev"` which is what I actually used.  However, some repositories have both 1.81 and 1.83, so perhaps it would be best to peg it at a particular version at this point, recognizing that this may change soon as repositories are updated.  I tried `1.8*` and in one instance there were errors due to both 1.81 and 1.83 being present, but with other repositories `1.8*` works fine.  As you point out, not all Debian derivatives are identical in their repositories, even though based on the same Debain version (e.g., Debian 12).

Maybe peg it at 1.81 and add a note to the effect that this version number may need to be changed depending on which most-recent version of Boost is in the repository. 

73,
-Joe-
K0OG

**Comment by Chris-AC9KH** [comment]
Well, boost wasn't the issue. Yes, the command failed because it wasn't found. But I fixed that. The bigger issue is that we're missing some Qt frameworks in those commands, in particular the SVG (Scalable Vector Graphics) library. I double checked and it did NOT install on my test system. I know it's included in the MacOS build because it's bundled and linked to the binary in the .app.

It's a runtime dependency - not required to build JS8Call, but required to package it. But not for a .deb, only for .rpm. So evidently Ubuntu is dynamically linking that framework, Fedora is statically linking it. Unless in Ubuntu it's part of one of those Qt6 meta-packages. It is available in Ubuntu 24 as libqt6svg6 but it did NOT install with the given commands until I explicitly installed it.

Sorry for the screenshots - I'm running Ubuntu in a VM so can't copy and paste back and forth between the two operating systems.

Screenshot 2025-08-20 at 07 15 49

Screenshot 2025-08-20 at 07 28 17

**Comment by Chris-AC9KH** [comment]
I'll add that without that SVG framework all you get on Ubuntu is the base ugly Gnome SAX-based render, you don't get the nice rounded-corner buttons etc that you get with the Qt SVG framework. Maybe this was done on purpose to make the Ubuntu UI butt-ugly? I'm not sure. It was either an oversight or it was intentional. Further investigation needed.

When I get time I'll create a new Qt Creator project that's up-to-date and do a little debugging in Qt Creator to see why this is happening.

**Comment by wmiler** [comment]
@Chris-AC9KH I hate to add more work for you, but... It might be useful to add in each of the build readme's a section of minimum supported libraries required.

**Comment by Chris-AC9KH** [comment]
> @Chris-AC9KH I hate to add more work for you, but... It might be useful to add in each of the build readme's a section of minimum supported libraries required.

The way it's looking, that is an absolute. In an .AppImage that Qt SVG framework would be bundled and statically linked. Not sure why there's a difference between .deb and .rpm. This might be why large projects that use Qt, like the dronecode QGC project, only build with Qt Creator and distribute via .AppImage.

**Comment by Chris-AC9KH** [comment]
It would be interesting to see some screenshots of what JS8Call looks like running on Ubuntu or Debian with Gnome, vs KDE vs Fedora. I'm pretty sure if you replace Gnome with KDE on Ubuntu that you get that SVG framework with KDE. What I'm seeing on Ubuntu/Gnome may have been intentional by the original designers of WSJT to retain the native "look and feel" of the desktop environment.

I'm thinking about taking a completely different approach to these linux builds where we pull Qt from git sources and build it instead of using distro-packaged Qt libs. It not only standardizes what version of Qt is used to build JS8Call, it might end up with a more favorable result vs chasing your tail because distro's package different versions of Qt all the time that might break things in the future. It appears we're already doing that with Hamlib, why not do it with Qt too? Qt is the big one that can break functionality of the program.

I would say the package maintainers building for releases can use whatever they want. But for end users who want to build for their platform, and scripted builds, we need consistent results. Just like on MacOS those libraries only need to be built once and you can re-use them indefinitely for subsequent builds of JS8Call until one day when you decide to make a change to one of the libraries. In other words, treat those libraries as submodules that are checked out at a specific version and they never change until you decide to upgrade one of them. Maybe now would the time to add them to the js8call project as git submodules?

**Comment by Joe-K0OG** [comment]
@Chris-AC9KH Yeah, building/installing a particular Qt version on Linux would standardize things, but is more work.  This is what is required for Windows builds.  Is this really necessary?  For basic functionality, probably not.  It becomes more important when distributing installers.  Don't we want to just provide the simplest basically functional build instructions here?  If someone wants perfection, they can pursue that on their own.

>Ultimately, going to Qt Creator is going to be better. There's already a js8call.pro file in the source tree, which is a project file for Qt Creator but it hasn't been updated since like 2011 and it's for the old Fortran build from WSJTx

I wasn't aware of this.  Doing this would be really nice if we can get it to work!  Fully automated, no need for build-tools knowledge :)

73,
-Joe-
K0OG

**Comment by Joe-K0OG** [comment]
@Chris-AC9KH Chris,

I just built a pristine Ubuntu 25.04 system, fully updated, and find that the procedure worked fine with modifications I sent, with the `libboost*1.83*-dev` install, and deleting `libhamlib-dev` from the list (it is not really necessary since Hamlib is separately installed in this procedure).  The resulting JS8Call installs and builds a .deb package fine, with one minor caveat:  I found on this system I needed to use the command `sudo cpack -G DEB` for packaging to work.

73,
-Joe-
K0OG

**Comment by wmiler** [comment]
@Chris-AC9KH So as a part of my CI work, and given my rapid multi-iterations trying to dig my way out of Win64 path issues, I've taken to building some library artifacts. I have hamlib 4.6.4 built under MSYS2, and are currently working on a QT 6.6.3 MSYS2 build. Note: we don't need all of qt6 for building js8call, you can call ./init-repository with just the modules you need (cuts building time as well).

As a bit of reference, building hamlib on the Github free runners takes around 25 minutes. Downside, I have to rebuilt the images every 7 seven days, as that is a branch restriction, although I understand if it was built on main, the retention is 90 days. So a different repo with the various libraries by system/version could be stood up if needed.

**Comment by Joe-K0OG** [comment]
FYI:  The default repositories on several distributions are:

* Ubuntu 25.04 - 1.83 and 1.88
* Debian 12 - 1.74, 1.81, and 1.83
* Mint 22.1 - 1.74 and 1.83
* Fedora 39 - 1.81
* Fedora 41 - 1.83

Since the common one is 1.83, maybe go with that at this time.  I doubt we need to worry about Fedora 39 since it's so far out, but builds do work fine on Fedora 39 with Boost 1.81.

73,
-Joe-
K0OG

**Comment by Chris-AC9KH** [comment]
> Downside, I have to rebuilt the images every 7 seven days, as that is a branch restriction, although I understand if it was built on main, the retention is 90 days. So a different repo with the various libraries by system/version could be stood up if needed.

That's unfortunate that the lifespan of a build is only 7 days on the free runner VM's. The one downside with git submodules is that when you pull them into your project with `git submodule update --init --recursive' they have to be built BEFORE the project build every time. Which is not a big deal on an Apple silicon Mac but it would slower than molasses in January for Qt on an Intel machine even with a Core i7 Super Turbo with Enhanced Rocket Boost.

But we could host the built libraries in a separate repo under js8call (maybe call it js8libs) and pull them as pre-built libraries. Put the source code for them in the js8libs repo, build the libraries and put them under a Release as tarball or zip downloads for each platform and architecture. This would require minimal maintenance, standardize the libraries, only ever has to be changed if we change a library version, which would then go in a new Release. I could see pre-built libs for:
- Windows x64
- Windows arm64 (eventually whenever Microsoft provides the build tools for it)
- Mac Intel (as long as Apple keeps supporting it)
- Mac arm64
- Linux x66
- Linux arm64

I'm really seeing distro-packaged libraries as a non-starter on linux due to no consistency between distro's even in the naming convention of the library.

**Comment by Chris-AC9KH** [comment]
> @Chris-AC9KH Yeah, building/installing a particular Qt version on Linux would standardize things, but is more work. This is what is required for Windows builds. Is this really necessary? For basic functionality, probably not. It becomes more important when distributing installers

@Joe-K0OG this is for scripted builds, and I'm arriving at the conclusion that build instructions for end users should follow pretty much the same context and methodology as used by the scripted builds. To do this successfully you need consistency, and need to eliminate GUI tools.

For production release or testing builds you can do it any way you want. On Mac, for instance, I've gone to building with Xcode because it properly signs every library and framework in the .app bundle with my Apple Developer Certificate where a normal cmake build doesn't. Xcode also handles linking and embedding of Frameworks independent of cmake configurations.

But for testing builds I'll use the scripted build system that Wyatt is building.

Before I go any further with this PR I think I'm going to create a js8libs repo on my GitHub account, populate it with the libraries and test it to see if it works the way I think it will. I might add the Qt, Hamlib and Boost libraries to it as submodules and just host the code for the FFTW library in the base repo since they don't do releases via GitHub like the other three.

**Comment by Chris-AC9KH** [comment]
> @Chris-AC9KH So as a part of my CI work, and given my rapid multi-iterations trying to dig my way out of Win64 path issues, I've taken to building some library artifacts. I have hamlib 4.6.4 built under MSYS2, and are currently working on a QT 6.6.3 MSYS2 build. Note: we don't need all of qt6 for building js8call, you can call ./init-repository with just the modules you need (cuts building time as well)

@wmiler I understand the problem. Been there, dealt with it, before. The nice thing about using git submodules is that the code in the libraries is locked-in at the time the submodule is created as part of the project. It never changes unless you decide to update one of the submodules.

This provides consistency and version control that we presently lack. Rather than having a MSYS2 build of the libraries stored on your local machine you should be able to fetch them from github on-demand, simply unpack the library tarball (or zip) in the desired location and run with it.

I'll let you know how my experiment with setting this up on my own repo goes. For doing scripted builds you can use something like:
if [! -d "/path/to/js8libs" ]; then
    curl -o js8libs.tar.gz https://github.com/path-to-downloadable-tarball
    tar -xzf js8libs.tar.gz ~/path-to-where-it-goes
fi

<execute-whatever-here-if-it-already-has-been-downloaded>

If you're doing this in the free runners and they delete it after a time, the script will just re-fetch it.

**Comment by Joe-K0OG** [comment]
If you are going for scripted builds, then indeed make it as detailed as needed to make them consistent.  That would be really nice if one could simply pull a script, run it, and let it fly while eating breakfast, to come back and find JS8Call nicely done.  Will such a scripting system be workable on low-specification machines (slower processors, low RAM, smaller storage)?  Minimum machine specifications per OS would be good to include.  I hope that approach works out - good approach!

73,
-Joe-
K0OG

P.S.  That kind of approach is what the JTSDK64-Tools for Windows is about, although it is more general than for one single application, but for a family of Hamlib-related applicatons (and more).

**Comment by Chris-AC9KH** [comment]
> Will such a scripting system be workable on low-specification machines (slower processors, low RAM, smaller storage)?

It will work fine because the heavy lifting is done on more capable hardware that builds the libraries in the first place. All your build machine has to do is fetch them and build JS8Call itself.

**Comment by aknrdureegaesr** [comment]
A small nit:

>       .Build hamlib in ~Downloads/hamlib with the following script (assuming the source has been obtained with git, and "git checkout" has been used to select the hamlib version, per above, but placed is ~Downloads/hamlib/src):
> ```
> cd ~Downloads/hamlib/src
> ...
> ```

You are missing a `/` here. It should really be `~/Downloads/hamlib` rsp. `~/Downloads/hamlib/src`. At least, `bash` doesn't like it without the `/`.

File: docs/BUILD-Linux.md [hunk 8]
```diff
@@ -0,0 +1,70 @@
...
+To obtain the JS8Call source and build, issue the following commands in a terminal window:
+```
+   `cd ~/Downloads \
+        && git clone https://github.com/js8call/js8call.git js8call
+```
+```
+    cd js8call && mkdir build && cd build \
+        && cmake -D CMAKE_PREFIX_PATH=~/Downloads/hamlib -D CMAKE_INSTALL_PREFIX=/opt/js8call .. \
+        && make -j 4 \
+        && sude make install
```
aknrdureegaesr: I think this is a typo and should read `sudo` instead of `sude`.

aknrdureegaesr: Also, putting `&&` at the end of the line rather than at the beginning of the next line removes the need for `\`. This isn't essential, there isn't anything wrong with the formatting as is, but I personally would consider it an improvement. Feel free to ignore if you think otherwise.

Chris-AC9KH: This whole commit is going to get reversed and start over as soon as I come up with suitable submodules.

wmiler:
> > @Chris-AC9KH So as a part of my CI work, and given my rapid multi-iterations trying to dig my way out of Win64 path issues, I've taken to building some library artifacts. I have hamlib 4.6.4 built under MSYS2, and are currently working on a QT 6.6.3 MSYS2 build. Note: we don't need all of qt6 for building js8call, you can call ./init-repository with just the modules you need (cuts building time as well)
> 
> @wmiler I understand the problem. Been there, dealt with it, before. The nice thing about using git submodules is that the code in the libraries is locked-in at the time the submodule is created as part of the project. It never changes unless you decide to update one of the submodules.
> 
> This provides consistency and version control that we presently lack. Rather than having a MSYS2 build of the libraries stored on your local machine you should be able to fetch them from github on-demand, simply unpack the library tarball (or zip) in the desired location and run with it.
> 
> I'll let you know how my experiment with setting this up on my own repo goes. For doing scripted builds you can use something like: if [! -d "/path/to/js8libs" ]; then curl -o js8libs.tar.gz https://github.com/path-to-downloadable-tarball tar -xzf js8libs.tar.gz ~/path-to-where-it-goes fi
> 
> If you're doing this in the free runners and they delete it after a time, the script will just re-fetch it.

So, I'm running a windows self-hosted runner. The builds are in my GitHub Action runs and stored on GitHub. I don't have the space to store them locally :) I call them currently from the CI yaml as follows:

``` yaml
    - name: Download hamlib-windows-msys2-binaries
      uses: actions/download-artifact@v5
      with:
        repository: wmiler/js8call
        run-id: 17105686888
        artifact-ids: 3811637899
        github-token: ${{ secrets.GITHUB_TOKEN }}
```

This works pretty well, and at some point, if I add it to a main branch repo, I can set it to run on a schedule, and it will autobuild in the future (not on my machine though, it's only a 4th gen i7).

Joe-K0OG:
@Chris-AC9KH Per my earlier comment, for building JS8Call I think it would be better to use cmake instead of make, thusly:

`cmake --build . -- -j 4`

73,
-Joe-
K0OG

Chris-AC9KH:
@Joe-K0OG I don't know that that provides any real advantage? cmake generates configuration files for other build tools based on what it reads in its configuration flat text file. The other tools are responsible for more directly building; using the generated headers. So whether it is cmake or make doing the actual build is not going to provide any performance improvement - that's all in the compiler.

Joe-K0OG:
@Chris-AC9KH I'm not proposing based on theory, but on experience.  I found that after doing a "make" then "cpack" it sometimes doesn't work properly, yet under identical conditions doing a "cmake" build yields a result that "cpack" works properly on.  I can't explain it, but just know "it works..."  ;)

-Joe-

docs/BUILD-Linux.md
aknrdureegaesr: The Linux build instructions for users that have a reason to build should not refer to any git submodules. For the simple reason that those are not needed.

For `hamlib` in particular, this has been discussed at some detail over at #41. Similar issues could be opened for the other libraries that JS8Call uses. But we currently do not have what I (disrespectfully) call "cargo cult", namely, not using what the distribution provides, but rolling our own. None of this is needed.

Indeed, I'm waiting for this PR to be merged to start my fix of #41.

For that, I intend to come up with a new PR that removes the entire `hamlib` build from the build instructions most people will follow. As I envision the fix for #41 , the `hamlib` build instruction will be moved to a second-class seat, somewhere at the bottom of the build instructions, meant for those people who need a new `hamlib` version to operate a radio that the new `hamlib` version supports, but the version that comes with the distribution doesn't, or if the support has significantly improved.

Most people will be fine simply using whatever libraries (hamlib, boost, QT6) their Linux distribution offers.

That, in my humble opinion, is what distributions are for: Provide a base of good-enough software that users do **not** need to compile themselves. Some unstable, actively developed software like JS8Call is an exception. But that does not make all the libraries such software needs an exception, too.

Joe-K0OG: >Most people will be fine simply using whatever libraries (hamlib, boost, QT6) their Linux distribution offers.

Some, most, maybe, but not all.  There is a constant stream of comments & requests on the forum for help with the latest rig which isn't showing up in the older Hamlib which is in the distribution (or perhaps has some bug which a more-recent version fixes).  It would at least be "polite" to have instructions for people to be able to provide their own build with the latest Hamlib.

73,
-Joe-
K0OG

Chris-AC9KH: > Most people will be fine simply using whatever libraries (hamlib, boost, QT6) their Linux distribution offers.

@aknrdureegaesr I think good in theory, but most people don't use linux. Windows is number 1 in desktop marketshare, MacOS is number 2. We need CI with automated builds that can run and test code whenever anybody submits a PR. Wyatt is working on a pretty important project there. To make this work we need consistency, not the hodge-podge that is linux distros where it's a crap shoot as to what library you're going to get, and what its name is. As I found out, the naming convention of the libraries in arm64 Ubuntu is not even consistent with the naming convention in x64 Ubuntu. 

This has been a problem forever in linux desktop systems, part of the reason why linux doesn't have much desktop marketshare, not even 5% globally. And in the U.S. MacOS rivals Windows for the number one spot. The majority of the current JS8Call codebase was written and developed on MacOS. So I think we need to do what's right to streamline this for the majority of our users, not a small subset.

If the linux guys want to mess around with building .deb's and .rpm's, you can still do that all you want. You don't have to use the submodules if you don't want to. They are there to streamline, simplify and standardize the build and CI scripting process for the main two desktop systems. When I first got involved with this project making sure I could get a build to work on MacOS and had several failures due to Qt versions, I'm like, why aren't we using submodules here instead of grabbing libraries from all over tarnation? Every other project I've ever worked on uses them.

Chris-AC9KH:
@Joe-K0OG @aknrdureegaesr I fixed a couple typos and mistakes in what's currently on this branch. There doesn't seem to be consensus on what commands should be use, cmake --build or make, etc. etc..

What I would suggest is if you guys could fork my js8call repo, check out the pr/build_documentation branch, make changes to it as necessary, submit a PR to that branch to be merged with your changes. When I merge your PR it will update the PR here. Then we can get this one out of WIP status for merge into project main.

The issue with submodules or what library versions, etc.. is really a separate issue from this one. It is more related to the CI scripted builds that @wmiler is working on. It will eventually streamline how builds are done. But it is non-trivial setting that up, and again, has nothing to do with this PR. I'm working on setting up a system that will make it easier for Wyatt to accomplish what he wants to with the scripted CI builds.

Joe-K0OG:
At this point I've contributed about all I can.  All I can say is what works for me on multiple platforms -- empirical solutions.

73,
-Joe-
K0OG

Chris-AC9KH:
> At this point I've contributed about all I can. All I can say is what works for me on multiple platforms -- empirical solutions.

@Joe-K0OG ok, when I get time I'll have to figure out why the commands given still omit the scalable vector graphics framework from Qt. JS8Call runs without it, but the UI doesn't render properly using the Gnome SAX-based renderer. At least not on arm64 Ubuntu. The drop down box that enumerates the audio interfaces doesn't even work.

I'll add that I did a full build of Qt6 from git sources last nite, built JS8Call with that and then it works. So either the Qt meta-packages from arm64 to x64 Ubuntu are different, or we're missing a Qt package there in the commands given.

Joe-K0OG:
@Chris-AC9KH Chris, when I set up my build systems using my original instructions, I wildcarded Qt and Boost installations so it installed much more than the specific list proposed by others.  I'm sure that would solve your problem -- go back to my original proposal.  Yeah, it installs more than actually needed, but it's no big deal since it is simpler, and you give it the apt install command and walk away for a few minutes.  After a cup of tea, it's done.  ;)

73,
-Joe-
K0OG

docs/BUILD-Linux.md
aknrdureegaesr: I'm not saying we should not **have** git submodules. If those help our own (and our users') Windows or MacOS builds, than so be it. I'm just saying we should not mention them in the standard build instructions **for Linux users** that want to build from source for any number of reasons

- to find out whether they can do it
- because they are running some CPU we don't yet supply a build for (like Raspi)
- because they want to build from a PR to evade a bug that's bugging them
- because they want to build from a PR to help test that PR's code
- because they want to build from a branch to develop a PR themselves

All of the above reasons typically don't need special versions of Hamlib. So we shouldn't make those people "pay for what they don't need" by making them compile Hamlib or any other of our libraries.

Then there is the additional reason

- People who want to compile themselves to make use of a new Hamlib that supports their rigs better.

To those people I would recommend: Try to combine the pre-built JS8Call with the new Hamlib's `rigctld`; that might be all they need to do. If someone actually wants to compile JS8Call with the new Hamlib (to use without intervening `rigctld`), that should be possible, too. For that use-case, let us preserve the existing Hamlib compilation instructions for Linux, but not as the typical thing everybody does, but only for those that have this requirement.

For the record: I don't have any Windows nor any Mac system, so remain rather quiet as far as those are concerned.

aknrdureegaesr: > If the linux guys want to mess around with building .deb's and .rpm's, you can still do that all you want. You don't have to use the submodules if you don't want to. They are there to streamline, simplify and standardize the build and CI scripting process for the main two desktop systems.

That is exactly what I'm after. I presently have no intention meddling with MacOS or Windows builds. If for those, git submodules with library versions are useful, than so be it. I myself am only (narrowly) interested in providing the easiest-to-use build instructions **for Linux users** wanting to build themselves. In those build instructions, I keep arguing we should do without git submodules.

Chris-AC9KH: > > If the linux guys want to mess around with building .deb's and .rpm's, you can still do that all you want. You don't have to use the submodules if you don't want to. They are there to streamline, simplify and standardize the build and CI scripting process for the main two desktop systems.
> 
> That is exactly what I'm after. I presently have no intention meddling with MacOS or Windows builds. If for those, git submodules with library versions are useful, than so be it. I myself am only (narrowly) interested in providing the easiest-to-use build instructions **for Linux users** wanting to build themselves. In those build instructions, I keep arguing we should do without git submodules.

Use of submodules is totally optional. I'm not changing the CMake configurations to use them. However, I do think they will be used for linux .AppImage where we want to bundle a specific set of libraries in the .AppImage. And after I get the submodules set up and working they will be used for MacOS builds vs fetching and building all the libraries as part of the build. I suspect they will be used for Windows builds as well with automated scripting to produce the .zip package for a Windows build.

The BUILD-Linux.md still needs some work, as it doesn't yet work on arm64 from Ubuntu ports for that architecture due to problems with Qt packaged from the distro. For one, I think they are using 6.4.2 instead of 6.6.3 and there appears to be problem there with the Qt submodules. It works when I build 6.6.3 from git sources on arm64, but not with the distro-packaged Qt 6.4.2. It might work on x86_64 but that doesn't do the raspi arm64 people any good until I figure out what the problem is.

aknrdureegaesr:
> What I would suggest is if you guys could fork my js8call repo, check out the pr/build_documentation branch, make changes to it as necessary, submit a PR to that branch to be merged with your changes. When I merge your PR it will update the PR here. Then we can get this one out of WIP status for merge into project main.

I did that. First stab. I'm not fully pleased myself: I made some "**FIXME**" remarks in the build instruction file, and I also think `~/Download/hamlib` and `~/Download/js8call` are suboptimal choices. But I hope it is a further step in the right direction. See https://github.com/Chris-AC9KH/js8call/pull/1 for details until Chris merges.

I'll be busy with stuff, not sure I can do much more before August 26.

Chris-AC9KH:
@aknrdureegaesr I won't merge that right away. I'll wait until you make more updates to it to get it to where you are pleased with it. But it does look good for a start to get the linux build issues sorted out.

A few days will give me time to work on my submodule project for scripted builds. And to investigate arm64 Ubuntu some more to try to figure out why we're missing a Qt framework with the distro-packaged Qt that appears to get installed on x86_64. If someone with a x86_64 machine could do `dpkg -l | grep qt >> installed.txt` and get the output of that command to me it would help. Otherwise I may have to install x86_64 Ubuntu 24 in my VM to make some comparisons because using the wildcard to try to install everything related to Qt doesn't work either on arm64. Ubuntu ports either deliberately didn't package it on arm64, or the build on that submodule failed so it wasn't included. The only way I've gotten it to work so far is to build Qt6 from git sources using ver 6.6.3 instead of 6.4.2

Without looking up the build logs for Ubuntu ports it's hard to say which version of ARM they compiled it on. If they're using ARMv8 on a Rasberry Pi the compiler may not handle alignment in integer registers when building Qt6. Apple's M-series processors allow the argument to start in an odd-numbered xN register. The standard ABI requires it to begin in an even-numbered xN register so there's some things that will build on Apple silicon that won't on ARMv8.

Joe-K0OG:
>using the wildcard to try to install everything related to Qt doesn't work either on arm64

Try apt-get instead of apt (if you haven't already).

I'll attach the installed.txt file containing all the Qt6 components on my Debian 12 system which I used to build the .deb installers Jordan has been distributing.

[installed.txt](https://github.com/user-attachments/files/21923936/installed.txt)

73,
-Joe-
K0OG

aknrdureegaesr:
I did a build in an Ubuntu docker container. The image was from `docker pull ubuntu:noble`, so 24.04. I (mostly) followed my own build instructions as of https://github.com/aknrdureegaesr/js8call/blob/a8256becbb8583aa14f8a37d87f2fd3d5ff5d7d3/docs/BUILD-Linux.md , slightly adjusted for Docker.

In that docker container, after the build completed nicely without problems:

```
$ dpkg -l | grep qt
dpkg -l | grep qt
ii  libqt6concurrent6t64:amd64                     6.4.2+dfsg-21.1build5             amd64        Qt 6 concurrent module
ii  libqt6core6t64:amd64                           6.4.2+dfsg-21.1build5             amd64        Qt 6 core module
ii  libqt6dbus6t64:amd64                           6.4.2+dfsg-21.1build5             amd64        Qt 6 D-Bus module
ii  libqt6gui6t64:amd64                            6.4.2+dfsg-21.1build5             amd64        Qt 6 GUI module
ii  libqt6multimedia6:amd64                        6.4.2-11build3                    amd64        Qt 6 Multimedia library
ii  libqt6multimediaquick6:amd64                   6.4.2-11build3                    amd64        Qt 6 Multimedia Quick library
ii  libqt6multimediawidgets6:amd64                 6.4.2-11build3                    amd64        Qt 6 Multimedia Widgets library
ii  libqt6network6t64:amd64                        6.4.2+dfsg-21.1build5             amd64        Qt 6 network module
ii  libqt6opengl6t64:amd64                         6.4.2+dfsg-21.1build5             amd64        Qt 6 OpenGL module
ii  libqt6openglwidgets6t64:amd64                  6.4.2+dfsg-21.1build5             amd64        Qt 6 OpenGL widgets module
ii  libqt6printsupport6t64:amd64                   6.4.2+dfsg-21.1build5             amd64        Qt 6 print support module
ii  libqt6qml6:amd64                               6.4.2+dfsg-4build3                amd64        Qt 6 QML module
ii  libqt6qmlmodels6:amd64                         6.4.2+dfsg-4build3                amd64        Qt 6 QML Models library
ii  libqt6quick6:amd64                             6.4.2+dfsg-4build3                amd64        Qt 6 Quick library
ii  libqt6serialport6:amd64                        6.4.2-4build2                     amd64        Qt 6 serial port support library
ii  libqt6spatialaudio6:amd64                      6.4.2-11build3                    amd64        Qt 6 Spatial Audio library
ii  libqt6sql6-sqlite:amd64                        6.4.2+dfsg-21.1build5             amd64        Qt 6 SQLite 3 database driver
ii  libqt6sql6t64:amd64                            6.4.2+dfsg-21.1build5             amd64        Qt 6 SQL module
ii  libqt6test6t64:amd64                           6.4.2+dfsg-21.1build5             amd64        Qt 6 test module
ii  libqt6waylandclient6:amd64                     6.4.2-5build3                     amd64        Qt 6 Wayland Client library
ii  libqt6waylandcompositor6:amd64                 6.4.2-5build3                     amd64        Qt 6 Wayland Compositor library
ii  libqt6waylandeglclienthwintegration6:amd64     6.4.2-5build3                     amd64        Qt 6 Wayland WaylandEglClientHwIntegration library
ii  libqt6waylandeglcompositorhwintegration6:amd64 6.4.2-5build3                     amd64        Qt 6 Wayland WaylandEglCompositorHwIntegration library
ii  libqt6widgets6t64:amd64                        6.4.2+dfsg-21.1build5             amd64        Qt 6 widgets module
ii  libqt6wlshellintegration6:amd64                6.4.2-5build3                     amd64        Qt 6 Wayland ShellIntegration library
ii  libqt6xml6t64:amd64                            6.4.2+dfsg-21.1build5             amd64        Qt 6 XML module
ii  qt6-base-dev:amd64                             6.4.2+dfsg-21.1build5             amd64        Qt 6 base development files
ii  qt6-base-dev-tools                             6.4.2+dfsg-21.1build5             amd64        Qt 6 base development programs
ii  qt6-gtk-platformtheme:amd64                    6.4.2+dfsg-21.1build5             amd64        Qt 6 GTK+ 3 platform theme
ii  qt6-multimedia-dev:amd64                       6.4.2-11build3                    amd64        Qt 6 Multimedia - development files
ii  qt6-qpa-plugins:amd64                          6.4.2+dfsg-21.1build5             amd64        Qt 6 QPA plugins
ii  qt6-serialport-dev:amd64                       6.4.2-4build2                     amd64        Qt 6 serial port development files
ii  qt6-translations-l10n                          6.4.2-1                           all          translations for Qt 6
ii  qt6-wayland:amd64                              6.4.2-5build3                     amd64        Qt 6 Wayland platform plugin
```

What does `js8call` really load?

```
$ ldd js8call
ldd js8call
	linux-vdso.so.1 (0x00007ffc13d14000)
	libhamlib.so.4 => /lib/x86_64-linux-gnu/libhamlib.so.4 (0x00007fb5da6cf000)
	libfftw3f.so.3 => /lib/x86_64-linux-gnu/libfftw3f.so.3 (0x00007fb5da497000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fb5da3ae000)
	libQt6Widgets.so.6 => /lib/x86_64-linux-gnu/libQt6Widgets.so.6 (0x00007fb5d9c7a000)
	libQt6SerialPort.so.6 => /lib/x86_64-linux-gnu/libQt6SerialPort.so.6 (0x00007fb5dce57000)
	libQt6Multimedia.so.6 => /lib/x86_64-linux-gnu/libQt6Multimedia.so.6 (0x00007fb5d9ba3000)
	libQt6Gui.so.6 => /lib/x86_64-linux-gnu/libQt6Gui.so.6 (0x00007fb5d9472000)
	libQt6Network.so.6 => /lib/x86_64-linux-gnu/libQt6Network.so.6 (0x00007fb5d92db000)
	libQt6Core.so.6 => /lib/x86_64-linux-gnu/libQt6Core.so.6 (0x00007fb5d8dcf000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fb5d8b51000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fb5dce27000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fb5d893f000)
	libusb-1.0.so.0 => /lib/x86_64-linux-gnu/libusb-1.0.so.0 (0x00007fb5dce07000)
	libindiclient.so.1 => /lib/x86_64-linux-gnu/libindiclient.so.1 (0x00007fb5d88a7000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fb5dce7e000)
	libudev.so.1 => /lib/x86_64-linux-gnu/libudev.so.1 (0x00007fb5dcdd4000)
	libpulse.so.0 => /lib/x86_64-linux-gnu/libpulse.so.0 (0x00007fb5d8856000)
	libEGL.so.1 => /lib/x86_64-linux-gnu/libEGL.so.1 (0x00007fb5dcdc0000)
	libfontconfig.so.1 => /lib/x86_64-linux-gnu/libfontconfig.so.1 (0x00007fb5d8805000)
	libX11.so.6 => /lib/x86_64-linux-gnu/libX11.so.6 (0x00007fb5d86c8000)
	libglib-2.0.so.0 => /lib/x86_64-linux-gnu/libglib-2.0.so.0 (0x00007fb5d857f000)
	libQt6DBus.so.6 => /lib/x86_64-linux-gnu/libQt6DBus.so.6 (0x00007fb5d84ca000)
	libxkbcommon.so.0 => /lib/x86_64-linux-gnu/libxkbcommon.so.0 (0x00007fb5d8481000)
	libGLX.so.0 => /lib/x86_64-linux-gnu/libGLX.so.0 (0x00007fb5d844e000)
	libOpenGL.so.0 => /lib/x86_64-linux-gnu/libOpenGL.so.0 (0x00007fb5d8423000)
	libpng16.so.16 => /lib/x86_64-linux-gnu/libpng16.so.16 (0x00007fb5d83eb000)
	libharfbuzz.so.0 => /lib/x86_64-linux-gnu/libharfbuzz.so.0 (0x00007fb5d82de000)
	libmd4c.so.0 => /lib/x86_64-linux-gnu/libmd4c.so.0 (0x00007fb5dcdac000)
	libfreetype.so.6 => /lib/x86_64-linux-gnu/libfreetype.so.6 (0x00007fb5d8212000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fb5d81f6000)
	libgssapi_krb5.so.2 => /lib/x86_64-linux-gnu/libgssapi_krb5.so.2 (0x00007fb5d81a2000)
	libbrotlidec.so.1 => /lib/x86_64-linux-gnu/libbrotlidec.so.1 (0x00007fb5d8194000)
	libzstd.so.1 => /lib/x86_64-linux-gnu/libzstd.so.1 (0x00007fb5d80da000)
	libproxy.so.1 => /lib/x86_64-linux-gnu/libproxy.so.1 (0x00007fb5dcda5000)
	libicui18n.so.74 => /lib/x86_64-linux-gnu/libicui18n.so.74 (0x00007fb5d7d8c000)
	libicuuc.so.74 => /lib/x86_64-linux-gnu/libicuuc.so.74 (0x00007fb5d7b7f000)
	libdouble-conversion.so.3 => /lib/x86_64-linux-gnu/libdouble-conversion.so.3 (0x00007fb5d7b6a000)
	libb2.so.1 => /lib/x86_64-linux-gnu/libb2.so.1 (0x00007fb5d7b4c000)
	libpcre2-16.so.0 => /lib/x86_64-linux-gnu/libpcre2-16.so.0 (0x00007fb5d7ac0000)
	libnova-0.16.so.0 => /lib/x86_64-linux-gnu/libnova-0.16.so.0 (0x00007fb5d7740000)
	libcap.so.2 => /lib/x86_64-linux-gnu/libcap.so.2 (0x00007fb5d7733000)
	libpulsecommon-16.1.so => /usr/lib/x86_64-linux-gnu/pulseaudio/libpulsecommon-16.1.so (0x00007fb5d76b5000)
	libdbus-1.so.3 => /lib/x86_64-linux-gnu/libdbus-1.so.3 (0x00007fb5d7666000)
	libGLdispatch.so.0 => /lib/x86_64-linux-gnu/libGLdispatch.so.0 (0x00007fb5d75ae000)
	libexpat.so.1 => /lib/x86_64-linux-gnu/libexpat.so.1 (0x00007fb5d7580000)
	libxcb.so.1 => /lib/x86_64-linux-gnu/libxcb.so.1 (0x00007fb5d7557000)
	libpcre2-8.so.0 => /lib/x86_64-linux-gnu/libpcre2-8.so.0 (0x00007fb5d74bd000)
	libgraphite2.so.3 => /lib/x86_64-linux-gnu/libgraphite2.so.3 (0x00007fb5d7497000)
	libbz2.so.1.0 => /lib/x86_64-linux-gnu/libbz2.so.1.0 (0x00007fb5d7483000)
	libkrb5.so.3 => /lib/x86_64-linux-gnu/libkrb5.so.3 (0x00007fb5d73b8000)
	libk5crypto.so.3 => /lib/x86_64-linux-gnu/libk5crypto.so.3 (0x00007fb5d738c000)
	libcom_err.so.2 => /lib/x86_64-linux-gnu/libcom_err.so.2 (0x00007fb5d7386000)
	libkrb5support.so.0 => /lib/x86_64-linux-gnu/libkrb5support.so.0 (0x00007fb5d7379000)
	libbrotlicommon.so.1 => /lib/x86_64-linux-gnu/libbrotlicommon.so.1 (0x00007fb5d7356000)
	libpxbackend-1.0.so => /usr/lib/x86_64-linux-gnu/libproxy/libpxbackend-1.0.so (0x00007fb5d7346000)
	libgobject-2.0.so.0 => /lib/x86_64-linux-gnu/libgobject-2.0.so.0 (0x00007fb5d72e3000)
	libicudata.so.74 => /lib/x86_64-linux-gnu/libicudata.so.74 (0x00007fb5d5583000)
	libgomp.so.1 => /lib/x86_64-linux-gnu/libgomp.so.1 (0x00007fb5d552d000)
	libsndfile.so.1 => /lib/x86_64-linux-gnu/libsndfile.so.1 (0x00007fb5d54a5000)
	libX11-xcb.so.1 => /lib/x86_64-linux-gnu/libX11-xcb.so.1 (0x00007fb5d549e000)
	libsystemd.so.0 => /lib/x86_64-linux-gnu/libsystemd.so.0 (0x00007fb5d53be000)
	libasyncns.so.0 => /lib/x86_64-linux-gnu/libasyncns.so.0 (0x00007fb5d53b6000)
	libapparmor.so.1 => /lib/x86_64-linux-gnu/libapparmor.so.1 (0x00007fb5d53a2000)
	libXau.so.6 => /lib/x86_64-linux-gnu/libXau.so.6 (0x00007fb5d539c000)
	libXdmcp.so.6 => /lib/x86_64-linux-gnu/libXdmcp.so.6 (0x00007fb5d5392000)
	libkeyutils.so.1 => /lib/x86_64-linux-gnu/libkeyutils.so.1 (0x00007fb5d538b000)
	libresolv.so.2 => /lib/x86_64-linux-gnu/libresolv.so.2 (0x00007fb5d5378000)
	libcurl-gnutls.so.4 => /lib/x86_64-linux-gnu/libcurl-gnutls.so.4 (0x00007fb5d52bd000)
	libgio-2.0.so.0 => /lib/x86_64-linux-gnu/libgio-2.0.so.0 (0x00007fb5d50ed000)
	libduktape.so.207 => /lib/x86_64-linux-gnu/libduktape.so.207 (0x00007fb5d50a0000)
	libffi.so.8 => /lib/x86_64-linux-gnu/libffi.so.8 (0x00007fb5d5094000)
	libFLAC.so.12 => /lib/x86_64-linux-gnu/libFLAC.so.12 (0x00007fb5d5030000)
	libvorbis.so.0 => /lib/x86_64-linux-gnu/libvorbis.so.0 (0x00007fb5d5002000)
	libvorbisenc.so.2 => /lib/x86_64-linux-gnu/libvorbisenc.so.2 (0x00007fb5d4f57000)
	libopus.so.0 => /lib/x86_64-linux-gnu/libopus.so.0 (0x00007fb5d4ef6000)
	libogg.so.0 => /lib/x86_64-linux-gnu/libogg.so.0 (0x00007fb5d4eec000)
	libmpg123.so.0 => /lib/x86_64-linux-gnu/libmpg123.so.0 (0x00007fb5d4e90000)
	libmp3lame.so.0 => /lib/x86_64-linux-gnu/libmp3lame.so.0 (0x00007fb5d4e1a000)
	libgcrypt.so.20 => /lib/x86_64-linux-gnu/libgcrypt.so.20 (0x00007fb5d4cd2000)
	liblz4.so.1 => /lib/x86_64-linux-gnu/liblz4.so.1 (0x00007fb5d4cae000)
	liblzma.so.5 => /lib/x86_64-linux-gnu/liblzma.so.5 (0x00007fb5d4c7c000)
	libbsd.so.0 => /lib/x86_64-linux-gnu/libbsd.so.0 (0x00007fb5d4c66000)
	libnghttp2.so.14 => /lib/x86_64-linux-gnu/libnghttp2.so.14 (0x00007fb5d4c3b000)
	libidn2.so.0 => /lib/x86_64-linux-gnu/libidn2.so.0 (0x00007fb5d4c19000)
	librtmp.so.1 => /lib/x86_64-linux-gnu/librtmp.so.1 (0x00007fb5d4bf9000)
	libssh.so.4 => /lib/x86_64-linux-gnu/libssh.so.4 (0x00007fb5d4b88000)
	libpsl.so.5 => /lib/x86_64-linux-gnu/libpsl.so.5 (0x00007fb5d4b74000)
	libnettle.so.8 => /lib/x86_64-linux-gnu/libnettle.so.8 (0x00007fb5d4b1f000)
	libgnutls.so.30 => /lib/x86_64-linux-gnu/libgnutls.so.30 (0x00007fb5d4925000)
	libldap.so.2 => /lib/x86_64-linux-gnu/libldap.so.2 (0x00007fb5d48c7000)
	liblber.so.2 => /lib/x86_64-linux-gnu/liblber.so.2 (0x00007fb5d48b5000)
	libgmodule-2.0.so.0 => /lib/x86_64-linux-gnu/libgmodule-2.0.so.0 (0x00007fb5d48ae000)
	libmount.so.1 => /lib/x86_64-linux-gnu/libmount.so.1 (0x00007fb5d4861000)
	libselinux.so.1 => /lib/x86_64-linux-gnu/libselinux.so.1 (0x00007fb5d4834000)
	libgpg-error.so.0 => /lib/x86_64-linux-gnu/libgpg-error.so.0 (0x00007fb5d480f000)
	libmd.so.0 => /lib/x86_64-linux-gnu/libmd.so.0 (0x00007fb5d47fe000)
	libunistring.so.5 => /lib/x86_64-linux-gnu/libunistring.so.5 (0x00007fb5d4651000)
	libhogweed.so.6 => /lib/x86_64-linux-gnu/libhogweed.so.6 (0x00007fb5d4609000)
	libgmp.so.10 => /lib/x86_64-linux-gnu/libgmp.so.10 (0x00007fb5d4585000)
	libcrypto.so.3 => /lib/x86_64-linux-gnu/libcrypto.so.3 (0x00007fb5d4072000)
	libp11-kit.so.0 => /lib/x86_64-linux-gnu/libp11-kit.so.0 (0x00007fb5d3ecc000)
	libtasn1.so.6 => /lib/x86_64-linux-gnu/libtasn1.so.6 (0x00007fb5d3eb6000)
	libsasl2.so.2 => /lib/x86_64-linux-gnu/libsasl2.so.2 (0x00007fb5d3e9c000)
	libblkid.so.1 => /lib/x86_64-linux-gnu/libblkid.so.1 (0x00007fb5d3e61000)
```

Comparing this with what `js8call` loads on my plain Debian 12 machine (the part before `=>` only), the interesting thing is that Ubuntu generally loads more stuff than does Debian. Doing it this way, no QT-related difference shows up.

```
$ diff -u ldd_ubuntu_cooked ldd_debian_cooked 
--- ldd_ubuntu_cooked	2025-08-21 21:19:18.127905740 +0200
+++ ldd_debian_cooked	2025-08-21 21:18:13.495980759 +0200
@@ -1,77 +1,56 @@
-	libapparmor.so.1 
 	libasyncns.so.0 
 	libb2.so.1 
-	libblkid.so.1 
 	libbrotlicommon.so.1 
 	libbrotlidec.so.1 
 	libbsd.so.0 
-	libbz2.so.1.0 
 	libcap.so.2 
 	libcom_err.so.2 
-	libcrypto.so.3 
 	libc.so.6 
-	libcurl-gnutls.so.4 
 	libdbus-1.so.3 
 	libdouble-conversion.so.3 
-	libduktape.so.207 
 	libEGL.so.1 
 	libexpat.so.1 
-	libffi.so.8 
 	libfftw3f.so.3 
 	libFLAC.so.12 
 	libfontconfig.so.1 
 	libfreetype.so.6 
 	libgcc_s.so.1 
 	libgcrypt.so.20 
-	libgio-2.0.so.0 
 	libGLdispatch.so.0 
 	libglib-2.0.so.0 
 	libGLX.so.0 
-	libgmodule-2.0.so.0 
-	libgmp.so.10 
-	libgnutls.so.30 
-	libgobject-2.0.so.0 
 	libgomp.so.1 
 	libgpg-error.so.0 
 	libgraphite2.so.3 
 	libgssapi_krb5.so.2 
 	libhamlib.so.4 
 	libharfbuzz.so.0 
-	libhogweed.so.6 
-	libicudata.so.74 
-	libicui18n.so.74 
-	libicuuc.so.74 
-	libidn2.so.0 
+	libicudata.so.72 
+	libicui18n.so.72 
+	libicuuc.so.72 
 	libindiclient.so.1 
 	libk5crypto.so.3 
 	libkeyutils.so.1 
 	libkrb5.so.3 
 	libkrb5support.so.0 
-	liblber.so.2 
-	libldap.so.2 
 	liblz4.so.1 
 	liblzma.so.5 
 	libmd4c.so.0 
 	libmd.so.0 
-	libmount.so.1 
 	libmp3lame.so.0 
 	libmpg123.so.0 
 	libm.so.6 
-	libnettle.so.8 
-	libnghttp2.so.14 
 	libnova-0.16.so.0 
 	libogg.so.0 
 	libOpenGL.so.0 
 	libopus.so.0 
-	libp11-kit.so.0 
 	libpcre2-16.so.0 
 	libpcre2-8.so.0 
 	libpng16.so.16 
 	libproxy.so.1 
-	libpsl.so.5 
+	libpthread.so.0 
 	libpulsecommon-16.1.so 
 	libpulse.so.0 
-	libpxbackend-1.0.so 
 	libQt6Core.so.6 
 	libQt6DBus.so.6 
 	libQt6Gui.so.6 
@@ -80,16 +59,10 @@
 	libQt6SerialPort.so.6 
 	libQt6Widgets.so.6 
 	libresolv.so.2 
-	librtmp.so.1 
-	libsasl2.so.2 
-	libselinux.so.1 
 	libsndfile.so.1 
-	libssh.so.4 
 	libstdc++.so.6 
 	libsystemd.so.0 
-	libtasn1.so.6 
 	libudev.so.1 
-	libunistring.so.5 
 	libusb-1.0.so.0 
 	libvorbisenc.so.2 
 	libvorbis.so.0 
```

Chris-AC9KH:
@Joe-K0OG thanks, that helped. You have Qt packages not available on arm64. I learned that arm64 Ubuntu is compiled on and built for ARMv8 (Raspberry Pi). Code compiled on Apple M1 or newer runs fine on it, but Ubuntu ports can't compile some things like the full Qt6 superproject, while Apple M1 can.

So no sense to trying the "fix" the linux build instructions for arm64. I think I would recommend that arm64 linux users run the .AppImage instead, compiled on a Mac. I wonder if .AppImage works on Raspberry Pi? It should, I think. I did a little search on the internet and there's people trying to run .AppImage on RPi but it fails because the .AppImage was built for x86_64. But I see no reason why .AppImage built for arm64 wouldn't run on RPi. I can attempt to build one if we can find somebody who can test it.

Chris-AC9KH:
@aknrdureegaesr I think x86_64 is not a problem. On MacOS it's also interesting that two of the Qt frameworks that are installed in the MacOS bundle are not binary linked by the CMake build. But building with Xcode instead of CMake, those frameworks are linked and signed. I had intended to look into that too, but haven't gotten around to it yet.

wmiler:
> @Joe-K0OG thanks, that helped. You have Qt packages not available on arm64. I learned that arm64 Ubuntu is compiled on and built for ARMv8 (Raspberry Pi). Code compiled on Apple M1 or newer runs fine on it, but Ubuntu ports can't compile some things like the full Qt6 superproject, while Apple M1 can.
> 
> So no sense to trying the "fix" the linux build instructions for arm64. I think I would recommend that arm64 linux users run the .AppImage instead, compiled on a Mac. I wonder if .AppImage works on Raspberry Pi? It should, I think. I did a little search on the internet and there's people trying to run .AppImage on RPi but it fails because the .AppImage was built for x86_64. But I see no reason why .AppImage built for arm64 wouldn't run on RPi. I can attempt to build one if we can find somebody who can test it.

I have a friend that has a rpi400. Since I just did freshly sync builds, I'll shoot it over to him to try out.

Chris-AC9KH:
@wmiler Wyatt, I got the submodules working. Basically you clone the js8lib repo, run `git submodule update --init --recursive`. When that completes run the BUILD.sh script. Just walk away and come back later and it will have created a pre-built library archive (currently only for Mac builds). Upload that archive to the Release here:
https://github.com/Chris-AC9KH/js8lib/releases/tag/js8lib-2.3

Just fetch that archive, unpack it - it will unpack to a directory named `js8lib` (no `s` on the end). You can rename `js8lib` to anything you like in your build script. Done. No need to fetch libraries from all over, no need to use brew (which is not going to be consistent), no need to build libraries for a JS8Call build because it's already done. Uses a set and standardized version of libraries from the repo for the build.

I've currently only run it on Apple silicon as a test, haven't run it on Intel yet.

Maybe change your build-macos.yml to use this and see how it works. I tested it here with an automated build script and had a Mac arm64 build of JSCall 2.3.2-devel in less than 5 minutes from scratch, including fetching and unpacking the library archive, cloning the js8call repo and doing the build. And all I did is sip some coffee, sit there and watch it work.

After I run it on Intel and upload a pre-built archive for that you can also do Intel builds for Mac.

Chris-AC9KH:
@wmiler Wyatt, the js8lib repo is not 100% production ready yet. I have to finish the README on how to use it because it requires a specific directory structure to work and the repo has to be cloned into a subdirectory called `submodules`. I'm going update the BUILD-Mac.md to use this instead of having users build libraries to do a MacOS build, or possibly fetch the wrong libraries which will result in a broken build.

Using brew on Mac to install libs is a non-starter because they won't statically link in the .app bundle and the resulting build will run ONLY on the machine it was compiled on. Using the submodules will solve a lot of these problems. The js8lib-2.3_Darwin_arm64.tar.gz is production ready, however, and can be fetched for a MacOS build. That has been tested and it works.

I'm also going to enhance the BUILD.sh script for Mac to do a validation build of JS8Call to test a library package build to verify that a pre-built library package uploaded to the 2.3 Release is tested and works. Scripts to pre-build library packages for Windows and/or Linux .AppImage can be created and added later. The current BUILD.sh script, however, will also run on linux since is is a pseudo Unix-like system.

Note: the js8lib repo is ONLY for developers doing scripted builds to create pre-built libs from a standardized set of library sources. It is not intended for end users to mess with, although they can if they want and understand how to use it. However, updating or experimenting with library versions with git submodules (in future branches that may use a different version of Qt etc) is usually beyond the capabilities of most end users.

aknrdureegaesr:
> > @Joe-K0OG thanks, that helped. You have Qt packages not available on arm64. I learned that arm64 Ubuntu is compiled on and built for ARMv8 (Raspberry Pi)
> > ....
> > So no sense to trying the "fix" the linux build instructions for arm64. ...
> 
> I have a friend that has a rpi400. Since I just did freshly sync builds, I'll shoot it over to him to try out.

I dare to suggest that some distributions for RPi are better than others. If Ubuntu on RPi gives troubles, I recommend to use the original Raspi OS instead.

I had almost no problems building with the straight Linux build instructions that I had polished and @Chris-AC9KH just merged into this PR. What problems I did have: I used `qemu` with a virtual Raspi 3 with just 1 GB RAM, which made the build unstable with `-j 4` parallelization. Removing the `-j 4`, the build came through without problems (after hours, on my 10-year-old laptop).

Chris-AC9KH:
@aknrdureegaesr Andreas I ran this on arm64 Ubuntu 24 from a fresh install and everything goes perfectly right up to packing the .deb. CMake is still looking for the local instances of rigctl and rigctld. It otherwise installs and runs fine, which is the first time I've gotten a successful build on arm64 'out of the gate'.

Screenshot 2025-08-30 at 10 16 31

Chris-AC9KH:
@aknrdureegaesr forget that previous comment. I followed the instructions verbatim to make sure it works on arm64 (which would be a new thing). I forgot to check out this branch that the PR is on for the final stage. That was just plain stupid on my part, was looking at CMakeLists.txt and trying to figure out what the problem is. Then it dawned on me, I'm on the wrong branch! It works fine :-)

aknrdureegaesr:
Thanks, @Chris-AC9KH ! Makes my day to read this! 🙂

Chris-AC9KH:
@jsherer I think this one is ready to go. I still haven't gotten a successful build on arm64 Windows 11, but that's more due to lack of development tools on arm64 and Microsoft's own compiler is not arm64-compatible. They seem to be taking the stance that they're going to run x86_64 applications on arm64 with the "prism" emulator since the Windows-based laptops with Snapdragon processors are not big sellers. So I don't think that one is possible right now.

Chris-AC9KH:
This is merged to rc/2.3.2 for merge into main.

Removing the PR from main

PR #39, opened 2025-07-29
Logging time always needs UTC, fix suggested by Joe K0OG

### LINKED ISSUE #37: Logging + time zone issue
I log a QSO and the times that are inserted into the UI are local time. It should be UTC.

Workaround: I can correct the times myself before actually writing to the log. If I don't do that, the time zone entered into `js8call_log.adi` is also local time.

This is a Linux box, the box's clock actually runs UTC, but the environment variables are set appropriately for Germany, which is currently on UTC+2.

The JS8Call version is self-compiled from 902f9f8165, which is a bit later than v2.3.1.

Thanks for providing fine software!

aknrdureegaesr:
See https://github.com/js8call/js8call/issues/37#issuecomment-3132750504 .

Fixes #37.

### DIFF
--- a/logqso.ui
+++ b/logqso.ui
@@ -164,6 +164,9 @@
                <verstretch>0</verstretch>
               </sizepolicy>
              </property>
+             <property name="TimeSpec">
+              <enum>Qt::UTC</enum>
+             </property>
              <property name="displayFormat">
               <string>yyyy-MM-dd HH:mm:ss</string>
              </property>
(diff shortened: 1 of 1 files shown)


Joe-K0OG:
> setTimeSpec is deprecated in Qt6. The compiler will flag a warning to use setTimeZone() instead.

It's good until we go to Qt6.10.  See my comment here...  https://github.com/js8call/js8call/issues/37#issuecomment-3133447324

73,
-Joe-
K0OG

Chris-AC9KH:
This PR has been built, tested, running it for some time now. I recommend for merge to master

jsherer (approved):

PR #42, opened 2025-07-31
Point users to the new installer place, 

aknrdureegaesr:
lest they find the old one http://files.js8call.com/latest.html through the JS8Call guide, that still advertises it.


Chris-AC9KH:
Recommend merge to master. It updates the README, which is bundled with the production builds.

jsherer (approved):
Thanks for the help @aknrdureegaesr

PR #46, opened 2025-08-01
Fix handling of receive timestamps. In particular, fixes #45.

### LINKED ISSUE #45: ALL.TXT bogus timestamps for all info that's being received since 2.3.x.
With JS8Call of the 2.2.x series, the `ALL.TXT` timestamps were comprehensible. After the upgrade to 2.3.x, this is no longer the case.

* The timestamp for lines of the JS8 frequency lines are good.
* The timestamp for records what I have sent are still good.
* The timestamp for records showing what I have received are wrong.

Here is a small excerpt of my `ALL.TXT` of action I believe happened 2025-07-30 09:43 - 09:44 Z showing all three:

```
2025-07-30 09:43:00  Transmitting 14.078 MHz  JS8:  VY 73 TO LEEDS FROM 
2025-07-30 09:43:30  Transmitting 14.078 MHz  JS8:  BERLIN!
2025-07-30 09:44:11  14.078 MHz  JS8
2025-07-30 16:12:00 -2 -0.3  863 A  ViiI6OAr+xqF         3    G3RCE: SM0RWO HEARTBEAT SNR -16 
2025-07-30 09:44:26  14.078 MHz  JS8
2025-07-30 16:12:15-21  0.4 2091 A  OWIi-BRpAZy0         1    2E0EEY: DJ3EI  
2025-07-30 09:44:41  14.078 MHz  JS8
```

The "16:12" times are bogus.

aknrdureegaesr:
This fixes two issues:

* The timestamp handling was a bit intransparent.
    * The documentation how an `int` needs to be converted into a timestamp or reverse was sketchy and incomplete.
    * The code to convert a timestamp into an `int` was at one place.
    * The code to reverse this was at a completely different place.

This code itself was correct, though, but it was harder than necessary to find that out. So this first issue is a documentation or "clean code" issue only.

The second issue was the bug described at #45. It derived from an innocent-looking expression `2 - period`, but in that expression, `period` was an `unsigned`, so the result was some number close to 2**32 instead of the intended -13 or so.


Chris-AC9KH:
This has been built, tested and running it. Recommend for merge to master

jsherer (approved):
:+1:

PR #47, opened 2025-08-01
Make it easier to understand the mode letters in ALL.TXT.

aknrdureegaesr:
There is no public information known to me available which mode letters in `ALL.TXT` correspond to which mode. Of course, with the exception of the source code.

Extracting this information from the source code is somewhat hard. Certainly harder than it needs to be.

This pull request improves the situation.

It also improves overall code quality, as it moves towards "single source of truth" regarding which number corresponds to which mode.

### DIFF
--- a/decodedtext.cpp
+++ b/decodedtext.cpp
@@ -30,11 +30,11 @@ namespace
   {
     switch (submode)
     {
-      case 0:  return 'A';
-      case 1:  return 'B';
-      case 2:  return 'C';
-      case 4:  return 'E';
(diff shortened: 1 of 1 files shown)


Chris-AC9KH:
This has been built, testing, running it since Aug 2. Recommend for merge to master

jsherer (approved):
Good add.

aknrdureegaesr:
> Good add.

Thanks. It is my pleasure to finally contribute to this software that I have enjoyed for years.

PR #57, opened 2025-08-10
WIP - Feature/cicd

wmiler:
CI/CD pipeline for Ubuntu/Debian on amd64 and arm64 archs.


wmiler:
This is still a WIP, I accidently opened the PR early. AppImages are currently building on my fork for amd64 Ubuntu.

Chris-AC9KH:
On MacOS builds need to eventually go to Xcode for proper signing of the libraries and binary, and building for iOS. I have a working Xcode build system using a different compilation of Qt that doesn't use Open GL that I've tested on iOS 18, but haven't decided how to incorporate it into the JS8Call source tree yet. Xcode requires a completely different layout for linking, signing and embedding libraries and frameworks since the cmake build is no longer used. Generation of a proper info.plist is handled in Xcode, obsolete Resources like INSTALL and  INSTALL-WSJTX are not included in a Xcode build, obsolete binaries like rigctl and rigctld are not included in a Xcode build, etc..

It may require a fork that pulls from upstream to accomplish this for MacOS and iOS builds. I just haven't decided how to go about it yet without completely re-arranging the current JS8Call source tree. I'm waiting for backlogged PR's to get merged into master so all developers are working on the same page. Right now everybody is going in different directions, and there's already PR's for significant fixes that need to go into 2.3.2 so we can all work off upstream master (or main) instead of pulling PR's for our development branches.

jsherer:
Thanks! I have done free time coming up that will be used for 2.3.2

Great work!

Best,
Jordan / KN4CRD

jsherer:
I’m fine with that! Do you want to pull together a draft “contributors guide” that we can all formalize and agree on?

Best,
Jordan / KN4CRD

wmiler:
> > One big help in this regard, would be on setting what the minimum system requirements will be for 2.3.2+
> 
> Currently, for MacOS I'm building with Hamlib-4.6.4, boost-1.88.0, fftw-3.3.10, libusb-1.0.29 and Qt-6.8.1. I've standardized on those because they are proven to build all branches of 2.3.x and work with no issues. I've never been one to jump on the latest and greatest release of libraries and frameworks when everything works unless there's a compelling reason to make a change.
> 
> On Linux it's a bit harder due to the variation in what various distro's package in their repos. So I think the minimum requirements are going to have to set for linux in able to support the LTS versions of Ubuntu. MacOS can run with just about anything other than Qt-6.8.3 where it works on Apple silicon but no audio on Intel. I tested with higher versions of Qt on Mac and no audio on either architecture with 6.9.1. So Qt versions need to be very carefully selected and tested.
> 
> Seems to me @Joe-K0OG mentioned he has to build with an earlier version of Qt on Windows because of OmniRig being broken with 6.8.1.
> 
> So it's possible setting the system requirements might be platform-specific.

I don't recall where I saw it (a lot of rabbit holes yesterday), but I think it was in the JTSDK that said there was a bug that was keeping the Windows builds at QT 6.6.3. Most of the CI/CDs in this PR are at Hamlib 4.6.4, Boost 1.88.0, fftw 3.3.10, libusb 1.0.28 or 29, and the QTs are all over the place one is at 6.4.2, a few at 6.6.x, and one at 6.9.x.

.github/workflows/build-macos.yml
```diff
@@ -0,0 +1,74 @@
...
+      run: brew install hamlib
+
+#    - name: Install fftw
+#      run: brew install fftw
+
+    - name: Install Boost
+      run: brew install boost
+
+    - name: Install QT6
+      run: brew install qt
```
Chris-AC9KH, line 41: I'll just comment here that if you grab Qt with homebrew you're going to get 6.9.1, which I've tested and the audio class is broken.

wmiler: Ok, thank you. I'll figure another way.

wmiler:
> Indeed Qt is the difficult one. Other projects I work on, namely QGroundControl for Dronecode/PX4, that are built on Qt typically standardize on one version of Qt that works across all platforms. They do NOT build and distribute linux distro .deb or .rpm packages where library versions are up for grabs - the only package for linux is the .AppImage that has the tested version of Qt bundled with it. On Windows the application is built with QtCreator and Visual Studio. On Mac it is built with QtCreator and Xcode. On Linux it is built with QtCreator with the ./qgroundcontrol/tools/setup/ scripts to obtain the dependencies to build it. On Android it is built on Mac or Linux with QtCreator and (currently) JDK17
> 

So a lot of these workflows are based off of the instructions in  #35 with caveats ie building QT is massive and time consuming, so I cheated, or because the setup just doesn't work via GitHub Actions (jtsdk is menu driven). Now that AppImages appear to be working (further testing needed, prolly this weekend), the debs will go away when that is completed.

.github/workflows/build-macos.yml
Chris-AC9KH: I changed the BUILD-MacOS.md (it's in a PR that's not merged with master yet) to use git sources to build Qt. Just checkout the 6.8.1 branch, use the ./init-repository script to fetch the submodules. You only need to build the submodules that JS8Call requires and once they are built you do not need to keep re-building them, they can be used over and over for subsequent builds. This is where hosting the built Qt products in a separate repo as a submodule for building JS8Call would be very helpful and time-saving. Make sure to include the Qt license with it.

As I get more into the submodule project (this is quite involved) I will have more and better info on how to do it, but this is basically what homebrew (and linux package managers) are doing. My goal is to create a submodule system for the required libs that work (initially) for MacOS, then expand it to cover the other two platforms. These can all be hosted on github in, say, js8call-Qt-submodule repo, and the same with the other libraries. But it will require a major overhaul of the cmake system to build it with submodules.

Like I said, it is quite involved to accomplish, it may be next winter before I get it done :-)

Chris-AC9KH: I should mention that hosting submodules as separate repo's is quite common with large projects. We hosted a repo for a custom-built version of the NuttX RTOS at the ArduPilot project for better than 10 years as a submodule before we switched to ChibiOS. I think the upstream NuttX sources were at version 8 and we were still at version 1 when we finally retired it. But it provided a very stable and predictable submodule with an excellent and robust scheduler without always chasing upstream for the latest and greatest. When you got something that works, you don't fix it :-)

Joe-K0OG:
Chris @Chris-AC9KH I'm reluctant to back down to 6.4.2 since we have done significant testing with 6.6.3 and it is performing well.  I don't see any pressing reason to drop to 6.4.2 and possibly introduce new problems.

73,
-Joe-
K0OG

Chris-AC9KH:
@wmiler Qt-6.6.3 works fine for MacOS. I get Qt from git sources, check out the 6.6.3 branch (to be the same as Windows) , run the init-repository script that's included in the github Qt sources to fetch and initialize the submodules. Then run configure with `-prefix <path-to-libs> -submodules qtbase,qtimageformats,qtmultimedia,qtserialport,qtsvg`

It cuts the build time drastically to only build the submodules you need vs the whole thing. Only takes a few minutes on an Apple silicon machine. If you build it with the fastest Intel processor you can find it will take a half a day. But I don't think going forward we're going to too excited about x86_64. MacOS 26 is the last version of MacOS that supports running x86_64 apps on Apple silicon with Rosetta2 and this puts us 6 years down the road since the switch to Apple-designed chips. Developers will drop support for x86_64 just like they did for PowerPC back in the day.

Chris-AC9KH:
What you can do on linux is uninstall all distro-packaged Qt-related software. Then get the official Qt Maintenance Tool from Qt. You can install whatever version of Qt you want with that tool to build AppImages.

Joe-K0OG:
I suggest that we stick with the Qt6.4.2 which is in the standard Debain-derived repos.  This is working well, and we have no pressing need to change it.  There is a lot of extra work to bump it to another version on Debian, etc.

73,
-Joe-
K0OG

Chris-AC9KH:
@Joe-K0OG it's a Mac mini M1 and running arm64 Windows in the UTM virtual machine using the native processor architecture (not emulated). However, on the Windows end they use what they call the "prism" emulator that apparently uses JIT compiling of x86-64 blocks to arm64 instruction set to run x86_64 apps on arm64. So I wouldn't expect it to work right and there's no sense to trying to debug that one.

I just did find a native arm64 version of Visual Studio for it that was just released a month ago, touted by Microsoft as the "first ever" native arm64 version of Visual Studio. That should contain the necessary compiler. But unless command line builds are possible, there's no arm64 version of the JDSDK that I can find.

I'll try the git bash shell and see if I can figure that out. As far as I know, as long as I got a compiler and can figure out the Windows commands, I SHOULD be able to compile the libs and the JS8Call binary similar to the way it is done on MacOS. I don't see any reason that the dynamically linked libs would have to be scattered all over the filesystem.

It will be an interesting learning project to see if I can get it to run on arm64 Win 11. It's maybe not important now. But with the Intel ship dead in the water and sinking fast, I think arm64 Windows is going to become a "thing" in the future.

Joe-K0OG:
@Chris-AC9KH Chris, I'm curious.  Have you tried JS8Call 2.2.0 or 2.2.1-devel on that same setup?  The processing in 2.3.x is far more efficient than it was in 2.2.x, so many changes were made in the reworking of the code (by Allan @bazineta).  I wonder if some of those changes may have affected how it works on the ARM64 through the emulator?

73,
-Joe-
K0OG

Chris-AC9KH:
> @Chris-AC9KH My some what working runner is [here](https://github.com/wmiler/js8call/blob/main/.github/workflows/build-win64.yml) and might give you a route to build for Windows Arm builds. 

@wmiler ok, now that looks interesting. There shouldn't be any significant differences in x64 vs arm64 other than the compiler used and what platform it's compiled on. That gives me a general guideline on how to do a commandline build on arm64 Win. Thanks!

Microsoft hasn't apparently put much dev effort into arm64 yet. I think they're using Qualcom Snapdragon chips (obviously Apple probably isn't gonna sell 'em any) and their compiler might be optimized for that. We'll see.

bazineta:
> > @Chris-AC9KH Chris, I'm curious. Have you tried JS8Call 2.2.0 or 2.2.1-devel on that same setup?
> 
> @Joe-K0OG no, I have not tried that. I can try it sometime this weekend to see if there's any difference. But historically, architecture emulators have never been 100% foolproof, especially where graphics is concerned.

If you go this route, there are some early fixes to the Fortran code in the original 2.3.0 mega-PR, without which I suspect you may find that the spawned process won't stay up. Sooner would I vacuum the dust bunnies out from under the fridge than revisit that stuff, but you do you.

Joe-K0OG:
@Chris-AC9KH Chris, if you want to do some more testing, try a 64-bit build of 2.2.1-devel from my Dropbox:

https://www.dropbox.com/scl/fo/j0kiml3wmka1a7dm0f6wt/h?rlkey=v5r84o3uneuz6fyoidtwqko1z&st=6x0e97mi&dl=0

73,
-Joe-
K0OG

bazineta:
> @Joe-K0OG ok, so I was eating lunch and fiddling with stuff. I tried v2.2.0 and it runs on arm64 Windows 11 but it don't decode at all. Mainly because the extra js8.exe process, which is 32-bit x86, crashed and burned right after I started it up. Which Windows says "something didn't go as planned - maybe try restarting your device or running Windows Update".

Chris, given the problem description, you likely need this commit to the Fortran code:

https://github.com/js8call/js8call/commit/5e4f1d848564b6beb5908d84ce5e59a12c49afc2

Absent that, the `js8` helper will be unlikely to work.

.github/workflows/build-macos.yml
wmiler: I'll look at your MacOS instruction updates today. I have zero problems going the submodule route for pinning the projects external libraries. That will help a lot. 

Windows update: GitHub recently (last 2 days) made several changes in their runner images that broke a bunch of stuff for a lot of projects. Currently stalled until they get things sorted. I'm currently trying two different routes to build, first is the MinGW and the second is completely without. We will see how the second route plays out.

I've actually used ArduPilot before, never built it from scratch however. Programming isn't my first skill, transmitters are however and sometimes you have to port antique code. 🤦‍♂️

Chris-AC9KH: > I'll look at your MacOS instruction updates today. I have zero problems going the submodule route for pinning the projects external libraries. That will help a lot.

On GitHub submodules have to be pulled from git sources. That works for every library except fftw. Not sure of a way around that one unless you would grab the fftw source tarball, unpack it and upload it to github in a submodule repo separate from the js8call one. The code that fftw hosts on their github repo is not buildable code, it is just a code generator.

Joe-K0OG:
I may misunderstand, but it appears that there is a push to eliminate .deb packaging for Linux builds.  I object (if that is the intent) to such a move.  It is good to keep packaging documentation in place for those who wish to generate packages for their own use.  I also argue that .deb packages continue to be provided for official distributions.  It is helpful to install .deb packages on low-memory and slower systems.  The AppImage format is for convenience, and works well on non-minimal systems.  Forgive me if I misunderstand something here.

73,
-Joe-
K0OG

wmiler:
> I may misunderstand, but it appears that there is a push to eliminate .deb packaging for Linux builds. I object (if that is the intent) to such a move. It is good to keep packaging documentation in place for those who wish to generate packages for their own use. I also argue that .deb packages continue to be provided for official distributions. It is helpful to install .deb packages on low-memory and slower systems. The AppImage format is for convenience, and works well on non-minimal systems. Forgive me if I misunderstand something here.
> 
> 73, -Joe- K0OG

Well, I was under the impression that the deb packaging caused too many problems (ie libc mismatch). I can revert the change if need be.

Chris-AC9KH:
I think the basic premise is that most people install their linux software from their distro repositories, and for Debian and Debian derivatives that's .deb. They typically build from source when the package maintainers build those packages. They'll grab a snapshot of the source code from the current stable release branch, build it and package it.

Whether is be .deb or .AppImage doesn't make one bit of difference if the system is minimal or non-minimal. They both require the same libraries, carry the same binary, consume the same system resources. Just that one is packaged without the libs, the other they are built-in.

Meaning that the one without the libs packaged depends on getting those from the distro repo. Even on Ubuntu 24 that fails with the .deb without doing a force update of the depends - I tried it on an install that previously had 2.2.0 packaged from the distro repo. I installed the .deb with dpkg -i and it failed to run. The .AppImage works out of the box.

The problem is that the current .AppImage was compiled on a system with a newer glibc. When you build an .AppImage it has to be built on the oldest version of glibc that you wish to support. At present I believe 2.3.5 is what the no longer supported Ubuntu 22.04LTS has, but there's still a lot of people running it anyway. If you build on that, you'll get minimum required 2.3.4. If you build on Ubuntu 24, which has glibc 2.3.9 then you're going get minimum required 2.3.8.

So you have to make that decision for the distributed package and publish what the minimum system requirements are for it.

Joe-K0OG:
Yes, please revert - it is good to have both options for .deb and AppImage.

73,
-Joe-
K0OG

Chris-AC9KH:
@wmiler Wyatt I got a start on the js8lib submodules tonight. At present I only have the Qt6 submodule installed, set to track 6.6.3. That was the big one and it only contains the Qt submodules, and the Qt dependancies required for those to build. The rest of the submodules will be fairly easy.

Screenshot 2025-08-20 at 22 05 08

I had to host the FFTW library as the base of the repo because FFTW only has code generating tools for developers on their github repo.

I think when I get this done it will simplify the CI scripting process by hosting pre-built libraries for various platforms and cpu architectures on this repo as a Release. The code here is NOT intended for people to build in order to build JS8Call, although it can be used for that. It is intended for developers to create pre-built and packaged libraries that can be used over and over again for JS8Call builds, particularly for Windows, MacOS, Linux .AppImage and retrieving libraries for scripted CI builds.

- The git submodules could be integrated into the base js8call project.
- Or they could be hosted in a separate js8lib repository on the js8call project later by simply forking this branch when I get it done.
- If there is no desire to integrate it into the js8call project I'll host it on my repo as a tool that developers (and end users) can use to build JS8Call
- It will standardize the libraries used (except for the linux guys who live and breathe .deb's and .rpm's and try to use whatever libs are packaged in their distro at this particular point in history).
- Eliminate having to fetch the library sources from all over tarnation.
- With submodules all it takes to make the switch to a new library version is to update the submodule and rebuild it.
- And (when I get it done) a pre-built library package can be created automatically for a particular platform and architecture with scripting, only having to fetch it from the Release to build JS8Call.
- It will ultimately save a lot time, because building the libraries is the time-consuming part.
- If i end up hosting it on my github repo permanently, I'll add any developers interested in contributing pre-built library packages as collaborators with write access.

https://github.com/Chris-AC9KH/js8lib

wmiler:
@Chris-AC9KH The [Hamlib msys2 CI](https://github.com/wmiler/js8call/blob/build-artifacts/.github/workflows/hamlib-msys.yml) needs to go into your .github/workflows directory, and can be adjusted as need to pull from your repo. Currently it pulls directly from Hamlib/hamlib 4.6.4, builds on the free runner and I think it's production ready. I'm running a build as we speak to confirm. Please note, the zip file that is produced, does not have an upper level directory ie no hamlib/include, just ./include etc.

Chris-AC9KH:
@wmiler ok, as I get closer to getting this ready for production use, test it and make sure it works, we'll add that.

wmiler:
Woot! I have a working Windows build! I've got a touch more work to do on the CI script, but should have it posted here Friday or Saturday some time.

@Chris-AC9KH I also have a working CI script for [QT 6.6.3 Windows msys2](https://github.com/wmiler/js8call/blob/build-artifacts/.github/workflows/qt663-msys.yml)

For those that want to exercise the [Windows ](https://github.com/wmiler/js8call/actions/runs/17141739808/artifacts/3824054910) build.

wmiler:
Just realized several things were merged recently to main, fresh builds for people that want to try them.
[Windows installer w/QT 6.6.3](https://github.com/wmiler/js8call/actions/runs/17143349167)
[Ubuntu 24.04 amd64 and arm64 AppImage and debs](https://github.com/wmiler/js8call/actions/runs/17143349176)
[Apple app w/QT 6.8.1 Arm64](https://github.com/wmiler/js8call/actions/runs/17144664833)

Chris-AC9KH:
@wmiler Wyatt, the MacOS build is not valid. The executable is not the right size, it's damaged and doesn't run even if I bundle it in a proper MacOS .app bundle with Xcode. Was this actually compiled on Apple silicon?

The zip just uncompresses to a Contents folder, not a .app. And it contains none of the libraries or frameworks required for a MacOS .app

Shown below on the left is what's in the downloaded file. Shown below on the right is what it is in a valid MacOS .app bundle. Although rigctl-local and rigctld-local are legacy and don't need to be there. Those are being removed in my Xcode builds. For whatever reason I haven't figured out yet, a CMake build installs the QtSvg and QtDBus frameworks but doesn't link them to the binary. An XCode build links those two frameworks to the binary and signs them. In either case the .app won't run without them. My gut instinct tells me XCode is doing it right, the CMake configuration in js8call needs a major overhaul.

Screenshot 2025-08-22 at 23 49 37

.github/workflows/build-macos.yml
```diff
@@ -0,0 +1,74 @@
...
+        make \
+        && make install \
+        && make clean
+
+    - name: Build JS8Call 
+      run: |
+        git clone https://github.com/js8call/js8call.git ~/development/js8call
+        mkdir -p ~/development/js8call/build && cd ~/development/js8call/build
+        cmake -DCMAKE_PREFIX_PATH="/Users/runner/development/js8libs;/opt/homebrew" ..
+        make -j 4
```
Chris-AC9KH, line 67: homebrew installs to /usr/local/opt but it won't statically link anyway because there is no binaries there, only symlinks. Using homebrew not recommended for libraries to build a MacOS .app bundle. Either build them from scratch (requires a specific directory structure in js8libs similar to AppDir on linux), or grab the pre-built libs on my js8lib repo, unpack those and use them. They already have the correct directory structure and were built on Apple silicon, not ARMv8 which uses a different instruction set.

leave the -j 4 off make. That doesn't do a darn thing on Apple silicon. An Apple M1 or later will run 20 or 30 parallel jobs by default.

You aren't running make install here, which is what actually builds the .app bundle. js8call.app will be found in the build folder after running make install.

wmiler: Yes, you are correct! Some how I completely left out the make install 🤦‍♂️
This is the [macOS runner](https://github.com/actions/runner-images/blob/macos-15-arm64/20250818.2206/images/macos/macos-15-arm64-Readme.md) I am using.

I am slowly getting rid of the the homebrew stuff, the macOS build isn't a priority for me at the moment, since I have no hardware to directly test on, and I'm not sure Js8call will run on my 1st gen Intel Mac. 🙄 Now that you have a release, I'll switch over to your library builds on my next run for macOS. 

Since you are working on a full Xcode build, let me know when you have the project file ready, and I'll switch over to that, as the runner image has Xcode installed.

Chris-AC9KH:
@wmiler I now have a fully functional library build system for JS8Call. It uses an optimized version of Qt6.6.3 as a submodule that only takes a fraction of the time to build. I can switch to whatever version of Qt I want to use by simply pointing the Qt6 submodule to a different branch. I locked in Qt6 sources from getting broken by upstream, since the Qt guys have a long and distinguished history of randomly breaking stuff and turning a currently usable release into an affair that has more bugs than a warm August night.
https://github.com/Chris-AC9KH/js8lib

You can also fetch the pre-built library archives for Intel and arm64 Mac from the Release, which were built with this system.

The BUILD.sh script fully automates the build, primarily intended for Mac, but also tested on arm64 Ubuntu 24 and it worked fine. While I have no intentions of running linux for anything useful other than testing stuff, you could create a Linux build script that would build the proper AppDir to make an .AppImage.

.github/workflows/build-macos.yml
Chris-AC9KH: > Since you are working on a full Xcode build, let me know when you have the project file ready, and I'll switch over to that, as the runner image has Xcode installed.

@wmiler I've been using Xcode builds for awhile. But this is not portable with a C++ project because the Xcode project uses absolute paths for the linker, etc.. If you want to use Xcode you first have to add the line `Xcode/` to .gitignore in every branch you have, including main. Any new branches you create from that point forward from will have the updated .gitignore entry. Then create a folder inside the source tree called `Xcode` At the command line cd into Xcode and run `cmake -G Xcode -DCMAKE_PREFIX_PATH=/path/to/libs and it will generate the Xcode project file.

It's not even close to functional. It will build a product off the bat, but nothing will be linked in it. It's just a debug build. You have to set Xcode up to build a release, which is quite involved and beyond the scope of explaining it here. For a release Xcode puts a lot of extra stuff in the Info.plist - it generates that file instead of cmake doing it anymore. And it has SDK's for deployment and testing on iOS (which, I've gotten JS8Call built and running on my iPad Air 5).

So this should be able to be done with commandline xcode tools on your runners. But the main advantage of using Xcode is the GUI and the extra developer tools it has for iOS (or iPadOS - I don't see JS8Call running on an iPhone).

And to get it running on iPadOS it requires a specially-built version of Qt for iOS (built from the same sources) that doesn't use OpenGL. OpenGL is not supported on iOS, and technically it's not supported on Mac although the SDK is still in Xcode for it. But I suspect that will be deprecated in the new version of Xcode for MacOS 26.

Chris-AC9KH:
After a couple bug fixes I think we have the pre-built library submodules working for MacOS. I did a build on arm64 Ubuntu 24.04.3LTS using the MacOS methodology to do the build and it worked. Although there's a couple extra Qt libraries needed for linux that MacOS doesn't need due to MacOS linking the binary with built-in System frameworks that linux doesn't have.

Since JS8Call links IOKit on MacOS, the app can't be put on the App Store. It's a public framework but Apple frowns on developers using it. For whatever reason I haven't looked into yet, using the CMake Xcode generator sets it in the Linker Flags and JS8Call won't build without it.

Screenshot 2025-08-26 at 22 27 27

I don't know enough about Windows to even go there.

wmiler:
> After a couple bug fixes I think we have the pre-built library submodules working for MacOS. I did a build on arm64 Ubuntu 24.04.3LTS using the MacOS methodology to do the build and it worked. Although there's a couple extra Qt libraries needed for linux that MacOS doesn't need due to MacOS linking the binary with built-in System frameworks that linux doesn't have.

With any luck, I'll have time to run another Mac build today that fixes the app not packaged as an app today. 

The particulars of the Windows builds on the runners have driven me nearly mad trying to get a 100% reliable and fast build, stable. As it is, and solely because of the OmniRig requirement, I have to build Js8call on my own self-hosted runner. 

I wonder if it might be a reasonable request, that we add a switch to disable the requirement for OmniRig to CMake. Please note, I'm not asking for it to be removed for release builds so long as people are still using it.

> 
> Since JS8Call links IOKit on MacOS, the app can't be put on the App Store. It's a public framework but Apple frowns on developers using it. For whatever reason I haven't looked into yet, using the CMake Xcode generator sets it in the Linker Flags and JS8Call won't build without it.
> 
> I don't know enough about Windows to even go there.

So from a brief look thru, it looks like the Mac build will need to move from IOKit to DriverKit to resolve that issue. Someone will need to port the serial and audio driver portion. DriverKit is supported on MacOS 12+. 

On the plus side, the move theoretically would allow an iOS port.  I would suggest opening a separate tracking issue for doing the DriverKit change.

Joe-K0OG:
@wmiler Wyatt, I don't really understand what you are trying to do in reference to a Windows build, but it's fairly straightforward to use the JTSDK64-Tools system, as it takes care of setting up all the tools for you.  Ignore this comment if it makes no sense... ;)

73,
-Joe-
K0OG

Chris-AC9KH:
There's also another issue I noted in the CMake build for MacOS. It is linking libraries and frameworks with `@executable_path/../Frameworks` when it should be using `@rpath` (or `@loader_path` works). This is not a problem with Xcode builds because the linker paths are set in Xcode and you script the `install_name_tool` to set the install id. But it does need to be fixed for a CMake configured build or it will stop working in MacOS 26.

Edit: and I'll note that the install id is set correctly to `@rpath` in the pre-built MacOS submodules. But when a CMake build is done it resets the install id to `@executable_path`.

wmiler:
> @wmiler Wyatt, I don't really understand what you are trying to do in reference to a Windows build, but it's fairly straightforward to use the JTSDK64-Tools system, as it takes care of setting up all the tools for you. Ignore this comment if it makes no sense... ;)
> 
> 73, -Joe- K0OG

Hey @Joe-K0OG nothing against the JTsdk system at all. But from a CI standpoint, totally useless :) 

From the CI standpoint, everything is scripted, running off somebody else's machine (GitHub/M$ Azure) and will be triggered every time somebody makes a pull request. As such, no human interaction allowed. Installing OmniRig **requires** human interaction, haven't found a way around that yet.

Chris-AC9KH:
> I wonder if it might be a reasonable request, that we add a switch to disable the requirement for OmniRig to CMake. Please note, I'm not asking for it to be removed for release builds so long as people are still using it.

I don't see where that would be a problem. CI builds are to check the code to make sure it compiles when changes are made. If a PR don't pass the CI build it has to get fixed before it's merged. This really has nothing to do with OmniRig or release builds.

Chris-AC9KH:
@wmiler I'll see you on js8lib and see if we can streamline the process for Windows.

bazineta:
> Since JS8Call links IOKit on MacOS, the app can't be put on the App Store. It's a public framework but Apple frowns on developers using it. For whatever reason I haven't looked into yet, using the CMake Xcode generator sets it in the Linker 

See the bit on the deployment target here; could be that setting a later one is what's required.

https://doc.qt.io/qt-6/macos.html#target-platforms

Additionally, later on in that same document:

_Applications can also be distributed through the Mac App Store. macdeployqt (bin/macdeployqt) can be used as a starting point for app store deployment. To ensure that Qt complies with the app store sandbox rules, Qt must be configured with the -feature-appstore-compliant argument._

Chris-AC9KH:
> See the bit on the deployment target here; could be that setting a later one is what's required.

That could be. I haven't looked into it yet since it's not the end of the world right now. Obviously an iOS app can only be distributed via the App Store, but that's not the case with MacOS (yet). It requires a completely different compilation of Qt6 to get JS8Call to run on an iPad. And then the UI has some issues and it's not exactly what I'd call usable.

Working on iOS and getting it to connect to wfview is a winter-time project for when it's 20 below outside and we only get 6 1/2 hrs of daylight for weeks on end :-)

Chris-AC9KH:
@wmiler I downloaded your latest MacOS CI build and it's broken. All the symlinks in the Qt frameworks have been removed, the libraries are not linked, the js8call (and rigctl-local and rigctld-local, which aren't actually needed) are all marked not-executable in the app bundle, as are the dylibs and Framework executables.

Screenshot 2025-08-27 at 19 06 14

So something is either going on with zip process that the runner is using when zipping it up (I HATE zip - use Unix Tape Archive). Or it's not being compiled on Apple silicon. Do you know what processor architecture your runner is actually using?

Another hint when building Mac .app's - you can't move them around at the command line with `cp` or you get this:

Working JS8Call.app - note the size of the .app bundle:
Screenshot 2025-08-27 at 19 15 06

After I moved it via the `cp -r` command at the commandline now look at the size of it
Screenshot 2025-08-27 at 19 18 32

I can tell by the size of the .app bundle in your latest build that it's been moved via `cp` or 'mv`, which may have broken the framework symlinks in it, possibly didn't preserve file permissions in it. If you are going to move a Mac .app bundle around at the command line either use Unix Tape Archive (tar) or Apple Disk Image (dmg). Do NOT use zip.

Chris-AC9KH:
The experimental build you did yesterday came with only the Contents folder, not bundled in a .app. But I dropped that one into Xcode and bundled it and it ran fine.

The one I tested today came in a `js8call` folder and the js8call.app was inside that. But something went awry in the process of moving the jscall.app into that folder. So I think your runner process is using zip, which it shouldn't do because not all zip programs are equal. And it's not recognizing that when it zips a js8call.app bundle that it's actually a directory.

This sounds like something Microsoft would do. And being this is github, which Microsoft bought, if they're doing this on a Windows platform it's gonna break it, guaranteed.

I guess the next thing to try is see if you can create a Apple disk image (dmg) before you let a brain-dead runner try to move it. They can zip up a .dmg and it SHOULDN'T hurt anything.

wmiler:
> The experimental build you did yesterday came with only the Contents folder, not bundled in a .app. But I dropped that one into Xcode and bundled it and it ran fine.
> 
> The one I tested today came in a `js8call` folder and the js8call.app was inside that. But something went awry in the process of moving the jscall.app into that folder. So I think your runner process is using zip, which it shouldn't do because not all zip programs are equal. And it's not recognizing that when it zips a js8call.app bundle that it's actually a directory.

Yeah, working on it, today I'll be out of the office, so not much time. Wish I could spin up a local Mac gui vm, but that's not technically legal... I'll shift the Mac build over to tar.gz on the next go around.

wmiler:
I would like to reiterate that this PR's intention is to be able to confirm that a PR will build WITHOUT error or new warnings. It really isn't for doing Official Public Release builds. If I can get developer PR build releases, that's great, but it's not a requirement.

Chris-AC9KH:
@wmiler I had to try this for myself to verify it. I have a JS8Call build from Xcode (Xcode produces `JS8Call.app`, where CMake produces `js8call.app` in lower case)

I zipped it using commandline zip and tarred and gzipped it using commandline tar. Then I unpack those archives.

JS8Call.app is the original

JS8Call 2.app is the unpacked tarball

JS8Call 3.app is the unpacked zip archive

Screenshot 2025-08-28 at 07 38 32

The Unix tape archive produces the original when unpacked. Zip doesn't. And you can also see that gzipped tar has superior compression. Depending on what flags you use with zip it won't retain the top level directory (which is JS8Call.app) and will only unzip to what's inside it (the Contents directory).

So I'm convinced you are getting a 100% successful build on your runner. But they are using zip to move it or upload and zip is breaking it. Unix Tape Archive (tar) is native to Unix systems like MacOS. Zip is not. There's too many variants of zip - 7zip, WinZip, WinRAR, etc.. There's only specification for tar and it hasn't changed since Unix Epoch. If you can get your runner to use tar instead of zip I'm pretty sure it will "fix it". Even though it's not a release build, it's still nice to test it to make sure it runs.

.github/workflows/build-macos.yml
wmiler: In progress, will follow up on your fork.

.github/workflows/build-macos.yml
wmiler: Starting on getting the MacOS build into shape today.

wmiler:
> > I wonder if it might be a reasonable request, that we add a switch to disable the requirement for OmniRig to CMake. Please note, I'm not asking for it to be removed for release builds so long as people are still using it.
> 
> I don't see where that would be a problem. CI builds are to check the code to make sure it compiles when changes are made. If a PR don't pass the CI build it has to get fixed before it's merged. This really has nothing to do with OmniRig or release builds.

This has been resolved. Switch no longer needed, see latest commit for the windows build. It's a bit hackish, but it does work, and PR building on my fork passes now.

wmiler:
I'm thinking I might close this PR and open a new one once @Chris-AC9KH and I get the MacOS build straightened out. There have been significant changes to the files which have not yet been committed on this PR branch, plus integrations with Chris's pre-built js8lib for the different platforms.

Chris-AC9KH:
@wmiler the commit history is quite varied right now on your work tree and that can probably be cleaned up using interactive rebase. But for the purpose of the PR I tend to agree - let's get it closer to "production status" and then submit a new one.

wmiler:
This PR is withdrawn, a new one will be along that targets just CI builds for all supported platforms.

PR #72, opened 2025-09-01
Correct MODE.SET_SPEED API

### LINKED ISSUE #50: API error, MODE.SET_SPEED
Problem reported via the groups.io forum, https://js8call.groups.io/g/main/message/26242, by KF7MIX:

_I notice than when I set the speed (MODE.SET_SPEED) with the API, JS8Call sets the speed in the Mode menu, but the WF red box does not change in width, the speed shown in the button at the top-right does not change, and most importantly the actual transmit speed doesn't change in the program._
 
_Is there a known bug in the API for this call?_

This does appear to have been unimplemented. The relevant code begins here:

https://github.com/js8call/js8call/blob/902f9f8165cecaa992c83c5c561adfa7a0ae43a8/mainwindow.cpp#L10699

And that section seems to be original, unchanged from this commit:

https://github.com/js8call/js8call/commit/a8d97d09b45fa2a32879faa4f087ae944ca30028

So yes, looks as if the problem is as described, the UI will appear to have had the speed changed, but in reality it's unchanged.

bazineta:
Adds setupJS8() call to MODE.SET_SPEED API implementation, causing the set speed to actually be set. Fixes #50.

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -10739,24 +10739,15 @@ void MainWindow::networkMessage(Message const &message)
     }
 
     if(type == "MODE.SET_SPEED"){
-        bool ok = false;
-        int speed = message.params().value("SPEED", QVariant(m_nSubMode)).toInt(&ok);
-        if(ok){
-            if(speed == Varicode::JS8CallNormal){
-                ui->actionModeJS8Normal->setChecked(true);
-            }
-            if(speed == Varicode::JS8CallFast){
-                ui->actionModeJS8Fast->setChecked(true);
-            }
-            if(speed == Varicode::JS8CallTurbo){
-                ui->actionModeJS8Turbo->setChecked(true);
-            }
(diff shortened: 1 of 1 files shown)


bazineta:
For a test harness, I've been using the [js8net](https://github.com/jfrancis42/js8net) package, which as of yet isn't installable via `pip`, so I just clone it locally. Test script:
```
import sys
sys.path.insert(0, '/Users/alb/Development/js8net')
from js8net import *
start_net("127.0.0.1", 2445)
set_speed(2)
```
Valid settings should return the speed set, e.g.:
```
>>> set_speed(2)
'2'
```
Invalid settings should return the current speed unchanged:
```
>>> set_speed(5)
'2'
```

Chris-AC9KH:
Notes: this merged to rc/2.3.2 for inclusion in the merge to main.

Conflicts resolved.

PR #84, opened 2025-10-03
Maintenance / Remove various files that no longer take part in the build. (in rc)

aknrdureegaesr:
To be precise, it is no longer linked into the executable.


Chris-AC9KH (commented):
Probably should remove the header file as well.

aknrdureegaesr:
This triggered my curiosity, so I whipped together a few lines of Python that analyze the dependency information the build generates, and checks whether each source file actually occurs somewhere in there. Well, there were a few source files that no longer did. Also, from earlier work, I know that JS8Call can be built without a FORTRAN compiler these days, so I conclude all FORTRAN source files are no longer needed, either.

I checked the streamlined repository still builds under Linux. It is an open question whether that is also true for Mac and Windows.

Chris-AC9KH:
@aknrdureegaesr This looks good for now. Probably a good idea to prefix your PR title with Docs/ (or in the this case) Maintenance/ to make it easier for Jordan to identify what category the PR is for when he gets around to reviewing them.

There is new CMake configs in the works that will deprecate much of the Fortran stuff, the CPack stuff, and move the minimum version of Qt required to 6.5 or greater.  We are now using qt_add_resources(), which is not supported on Qt 6.4. This does not affect Windows or Mac since both use newer Qt versions. But it does affect some linux distributions who still package Qt 6.4. Support ended for Qt 6.4 back in 2023. The linux distros affected (Linux Mint is one of them) will either have to use .AppImage or obtain Qt from the official sources at the Qt Group instead of using their distro packaging. It is a major overhaul of the build system for JS8Call and there's a limit to how much legacy stuff we can support.

The new CMake configs no longer call CPack - so packaging for distribution has to be handled post-build. This is done automatically on Mac and Linux with macdeployqt and windeployqt. Linux is not supported for this by the Qt Group. So we need a linux developer to tell us how this is going to work building on linux.

It is not under PR yet, still under development. You can clone my fix/cmake branch https://github.com/Chris-AC9KH/js8call/tree/fix/cmake and incorporate your changes into it to see how it builds.

aknrdureegaesr:
I saw your branch a while ago, checked it out, but the build recipe from the docs directory no longer worked for me. Didn't bother investigating what you're up to. Instead, I assume that some time in the future, a cleaned up version will appear in a pull request.

So I went back to a7baaee1060 as my personal "I'd which this was in `main`".

Chris-AC9KH:
@aknrdureegaesr I can confirm that this PR builds fine on MacOS with the new (upcoming) CMake configs. I can test it on Windows 11, but I suspect it will be fine there too.

Chris-AC9KH:
> Ah! And now I understand why id would not work for me. I'm using Debian `bookworm`, which comes with QT 6.4.

Yes, however, I've found if the distro doesn't have a new enough version of Qt, the sources.list can be modified and pull Qt 6.8.2 from Trixie. So it's not an unsurmountable problem. We just need a linux developer to explain to us how the post-built packaging is going to work without CPack. And to probably document what needs to be done on older distro's that won't have the required Qt version.

Chris-AC9KH:
@aknrdureegaesr unfortunately, OmniRig needs to stay for now. The new CMake configs have a switch to turn it off or on during build. But the configs are designed to remove it as the project can no longer officially support OmniRig. It is a DOA 32-bit application and compiling support for it into the code is the primary reason JS8Call gets flagged as containing a "trojan" by virus scanners on Windows 11.

So we need to leave it there for now, as probably the only way it can be supported going forward is for end-users to build their own JS8Call from source with the OmniRig switch turned on. Distributing JS8Call with it built-in is simply not going to work going forward.

aknrdureegaesr:
Ah! And now I understand why it would not work for me. I'm using Debian `bookworm`, which comes with Qt 6.4.

Personally, I intend to change that later this month, moving to Trixie, which comes with Qt version 6.8.2.

But I'm a bit worried to see support for Qt 6.4 dropped:

Microsoft want to earn money through its bulk license program, so Win 11 is not willing to run on a lot of perfectly serviceable computers, and Win 10 ends official support October 14. So Microsoft wants users owning such serviceable computers to ditch them and buy new computers, which invariably come with new bulk licenses from Microsoft, which makes Microsoft earn money.

Quite a few people doge that by moving over from Windows to Linux right now.

Linux Mint is a distro that has a good reputation for being easy to work with for people coming from Windows and just learning the ropes on Linux. The current 22.2 is LTS and will be supported until April 2029. I just did some homework and found out Mint 22.2 is based on Ubuntu 24.04, which is itself an obvious candidate for people to install these days. Ubuntu 24.04 is also a LTS version, with basic support until - what a surprise! - April 2029. Ubuntu 24.04 presently ships with Qt 6.4 as well.

So by forcing Qt 6.5+, we are hurting quite a few of our users, which will be pushed to app-images and such, instead of just being able to install JS8Call natively and be done with it.

In what sense is Qt 6.5+ so much better than Qt 6.4 that making life more complicated for our users is worth it?

Chris-AC9KH:
> In what sense is Qt 6.5+ so much better than Qt 6.4 that making life more complicated for our users is worth it?

Actually, Ubuntu 24 comes with Qt 6.8.3, which is the current "official" release in the 6.8-series from the Qt Group.

Linux is not linking the required libraries for the audio classes in Qt6 that are not supported in 6.4. Nor does 6.4 support qt_add_resources(). There was also significant changes to the IODevice handling in 6.5 and newer and too many linux folks complaining about PTT handling. Or JS8Call refuses to supply audio - it keys the rig but no audio. So they just go back to 2.2.0 because 2.3.1 don't work.

Mac and Windows don't have this problem.

If you've worked with Qt long enough, you know that Qt is a stickler this way. You can code your app to work with a particular version of Qt (6.8.3 is what the current codebase is developed on). With Qt there is technically no such thing as a MINIUM required version, it is a REQUIRED version (which should be 6.8.3). It will (probably) build on a different version of Qt, but there's no guarantee it won't inject phantom bugs. And that's where we're at.

The experienced and long-established Qt-based projects only allow one version of Qt, like this one that I also work on - scroll down and read about the required Qt version:

https://docs.qgroundcontrol.com/master/en/qgc-dev-guide/getting_started/

aknrdureegaesr:
This has been merged to https://github.com/Chris-AC9KH/js8call/tree/rc/2.3.2 and has been marked (in rc) to reflect that.

This PR can be unceremoniously closed when the above branched has been merged into this repo's `main`.

PR #91, opened 2025-10-15
Version 2.4.0-dev

Chris-AC9KH:
- contains a rollup of all outstanding PR's
- modernized CMake configurations
- bug fixes for audio, auto HB handling
- new features for group messaging
- new debug logging capabilities
- allow code signing with hardened runtime for MacOS
- updated build documentation for all three platforms
- developer guide
- general code cleanup and remove deprecated files
- updated license to be in compliance with Qt licensing requirements for re-distributed dynamically linked libraries
- set version to 2.4.0-dev


aknrdureegaesr:
I have some improvements to the build_for_Linux file at https://github.com/Chris-AC9KH/js8call/pull/6 .

It basically re-sorts the content as presented here, and removes material that can no longer be used to build the commits it is a part of.

aknrdureegaesr:
I found a bug in this branch:

People who build themselves expect to be able to run

```
cmake --install .
```

to install the binary `JS8Call` to the appropriate directory below `CMAKE_INSTALL_PREFIX`. This presently does not work.

aknrdureegaesr:
For a minor remark: I like the name `js8call` better. With this branch, the name of the binary built changes to `JS8Call`. Is that a bug or a feature? - It certainly isn't a big issue, I don't think.

But it might invalidate quite a bit of documentation that's out there.

Chris-AC9KH:
> For a minor remark: I like the name `js8call` better. With this branch, the name of the binary built changes to `JS8Call`. Is that a bug or a feature? - It certainly isn't a big issue, I don't think.
> 
> But it might invalidate quite a bit of documentation that's out there.

This is a feature. It makes no more sense to use `js8call` for the title of our program than it does for LibreOffice to use `libreoffice` for theirs. The project is named `JS8Call` on the official website. `JS` stands for Jordan Sherer, the inventor of the project. `JS8` is the mode, which was derived from FT8.

Chris-AC9KH:
> I found a bug in this branch:
> 
> People who build themselves expect to be able to run
> 
> ```
> cmake --install .
> ```
> 
> to install the binary `JS8Call` to the appropriate directory below `CMAKE_INSTALL_PREFIX`. This presently does not work.

It is not a bug. It is by design. The CMake install() command generates installation rules for a project in the build directory which can be used by CPack. The use of CPack for MacOS and Windows is deprecated due to better post-build tools for installation that create and sign the application's installer. This is required now, where we used to get away without doing it before. Plus CPack was not packaging the correct Qt dependencies.

So on Linux this post-build can be scripted. I have such a tool for MacOS that I considered including in the Darwin directory in the source tree. But I decided against it because the tool requires signing certificates from Apple be installed in the keychain on the machine doing the build.

But you can do a post-build script and install in the Debian directory if you want. Or, you can use CPack - the configuration files for it have been conveniently left in the Debian directory for you to use that if you want. You should find cpack installed in maybe /usr/bin or something on linux with your cmake install. In all cases we are not using include (CPack) anymore in the build configs because it yields a broken configuration on the two major platforms. And likely will on linux too with the new requirements for FFmpeg audio in Qt 6.8.3

This is also why we are attempting to standardize on Qt 6.8.3 - the Qt Group is deprecating use of native audio backends and we need to build and distribute a supportable configuration using a LTS version of Qt that's supported until 2029. To-date this has not been happening. Everybody has been going in different directions with the Windows guys wanting OmniRig (which is another issue that we can't officially support). The linux guys wanting to use distro-packaged Qt and ending up with complaints of PTT handling, audio dropouts or doesn't work, mysterious drops of decoding, etc..

So we are attempting to get the project in a supportable, consistent configuration. Which is part of the purpose of this branch to get a start on that. But it does require some changes that are not what was used in the past.

Chris-AC9KH:
> I have some improvements to the build_for_Linux file at [Chris-AC9KH#6](https://github.com/Chris-AC9KH/js8call/pull/6) .
> 
> It basically re-sorts the content as presented here, and removes material that can no longer be used to build the commits it is a part of.

I have not merged this PR (yet) for inclusion here. What I prepared was a rough draft that I figured would have to be updated once proper linux post-build is established. That has not been done yet. Remember, this merge is a dev branch, not a release branch. Which means it needs some more updating before we branch it to a 2.4-rc. The purpose here is to get all developers back on the same page so we can merge PR's without extensive conflict resolution, as it was not an easy project getting this put together with the new CMake configurations that improves build stability and consistency.

I did this to help @jsherer because I looked at it and it was an overwhelming ordeal to assemble it unless you have been tracking and actively testing commits that aren't even included in the outstanding PR's. I squashed many commits from Allans branch, which he and I collaborated on to produce working new CMake configurations. And it would've been virtually impossible for Jordan to figure that out.

So I would suggest holding off on any build instructions for linux until we all get back on the same page and a suitable post-build for linux is established. And I think a linux developer should be looking at that because I'm a Unix guy and we did not even have agreement on where the binary and libraries should be installed on linux. The default in Unix (and the CMake install() ) is /usr/local. So that needs to be sorted out first to make sure it is done the "linux way".

aknrdureegaesr:
> [JS8Call capitalization] is a feature.

Ok. Fine with me. - (We might want to mention that name improvement in our release notes.)

> It makes no more sense to use `js8call` for the title of our program than it does for LibreOffice to use `libreoffice` for theirs.

For your amusement: To this day, when I want to start LibreOffice, I type `libreoffice` into my command line. But what that uses is actually a symbolic link I came up with, from that name to `soffice`. "[StarOffice](https://en.wikipedia.org/wiki/StarOffice)" was predecessor of "OpenOffice" which in turn was a predecessor of "LibreOffice". The LibreOffice folks, in their official distribution, still call their entry point `soffice`, to this day.

I'm "old school" and like to type stuff at the command line, and I cringe a bit at program names that contain capitals. But old-school folks like myself know how to set up symbolic links, or shell aliases, or what have you - so this is not a concern.

aknrdureegaesr:
> [No `cmake --install .` ] is not a bug. It is by design.

I'm talking from the point of view of people who build themselves. And if you press me about it: Who build themselves on Linux.

The people I have in mind typically use the same machine for both building and running. So there is no need for those to come up with a signed package, which in turn typically requires a GPG setup they may not have. A plain "install" step could be helpful for those people, that simply copies the executable to whatever directory they originally wanted as `CMAKE_INSTALL_PREFIX`, stripping it in the process.

If we don't have such a install step yet, fine, then let those people run the binary unstripped where it sits after compilation.

No reason to not merge this as is.

jsherer:
Thank you for helping here. It was a bit daunting taking a look yesterday
:)

I'll review #91 today

Chris-AC9KH:
> > [No `cmake --install .` ] is not a bug. It is by design.
> 
> I'm talking from the point of view of people who build themselves. And if you press me about it: Who build themselves on Linux.
> 
> The people I have in mind typically use the same machine for both building and running. So there is no need for those to come up with a signed package, which in turn typically requires a GPG setup they may not have. A plain "install" step could be helpful for those people, that simply copies the executable to whatever directory they originally wanted as `CMAKE_INSTALL_PREFIX`, stripping it in the process.
> 
> If we don't have such a install step yet, fine, then let those people run the binary unstripped where it sits after compilation.
> 
> No reason to not merge this as is.

Well, I came up with a wild idea that I got from the ArduPilot project. I'm going to script an interactive install system for (initially) Debian 13 that fetches all dependencies and installs them, downloads the JS8Call source code, builds it and installs it complete with launcher icon in the menu.

The idea is that not all systems have the same dependencies available. There has been problems in the past with people getting .deb's to run due to dependency issues. The AppImage doesn't always run due to glibc problems. If the program builds on the target system, it will run every time.  I did a basic test run of this this morning, and it only took about 15 minutes or so to fetch everything, build it and install it on my arm64 Debian 13 system, with Qt 6.8.3

I need to script it so the installer detects the architecture and selects the proper Qt library. Debian does not add the user to sudo by default, need to check if that's the case because I refuse to let the script as root. Etc. etc.. After I get the basic scripting done it can be made as fancy or basic as you want.

Chris-AC9KH:
> > [JS8Call capitalization] is a feature.
> 
> Ok. Fine with me. - (We might want to mention that name improvement in our release notes.)

For right now, changing to upper/lower case for the project name is a nice feature. When I add an automated scripted build for my dev/2.4 branch it will allow linux users to build, run and test the branch easily without affecting their current JS8Call installation. The icons for both JS8Call programs can co-exist in the launcher/menu because one is lower case, the other is upper/lower case and can be titled JS8Call-dev.

This will allow linux users to easily install the latest code for testing instead of just MacOS. Unfortunately, I CAN do Windows but I don't want to. I find Windows extremely hard to work with because the command line environment is so convoluted and I don't even know how to do Windows batch files. And it's possible Windows 11 would not let batch files run anyway. I could set it up in Qt Creator for them, which would be easier. But it involves getting the proper libraries, which is easy on Mac/Linux with shell script. Not so easy on Windows.

Unfortunately, for the purpose of a scripted build, Ubuntu 22 cannot be supported. It does not contain a new enough version of boost to build JS8Call. It could be scripted to fetch the boost sources, and build it, which I did on the Ubuntu 22 server that I "borrowed" for testing. But it's not practical for the intended purpose. Debian 12/13, Ubuntu 24 and Mint should be easily doable though, with only minor changes to the scripting for each one to fetch the required libraries that are unique in each distribution.

PR #95, opened 2025-12-05
Take this repo out of production.

aknrdureegaesr:
Is this what we want?

Does this straight simple note suffice, or do we want more info?


aknrdureegaesr:
Explanation: This removes all files from the HEAD of the repo with the exception of the README.md, which is reduced to a very short note that redirects to the `js8call-improved` repo.

You can view that README.md here: https://github.com/js8call/js8call/blob/8a572e4d8f69b037aa1f323d4ac4ae7ed6431a66/README.md

Chris-AC9KH:
@aknrdureegaesr I don't feel that files from this repository should be removed. It should be preserved no matter what. I will fork it immediately to preserve it if this is going to happen.

jsherer:
It is not necessary to remove the repo or files. It is a historical artifact.

aknrdureegaesr:
I was not proposing to actually remove anything. The last repository content would have been one click away at https://github.com/js8call/js8call/tree/a7ff1be0b389d287fdc56e2ea0d06962aa68127d , even after my PR would have been merged. Maybe I should have mentioned that in the README.

A repo that looks like it is still active misleads people. I have seen time and effort wasted over and over again.

jsherer:
I’ve updated the readme and will be pointing things over to the improved repo.

Cheers!

Best,
Jordan / KN4CRD

Chris-AC9KH:
I like the way Jordan did it. I feel it is important to keep the branches, release(s) and any documentation intact because it is a historical record of the code, which no longer exists at bit bucket. Over at bit bucket I had to go thru a bunch of convoluted links to figure out how to pull up the source archive so I could grab the old Hamlib API out of it.

We link back to this repository in our documentation. It is a good reference to have. We have retained the 2.3.1 branch in our code, with the release tag for it. But other than searching all back thru the commits, that's as far back as we go for releases, and that code will no longer build on MacOS 26. So this is a very valuable historical reference because it contains Allan's merge branches in the Fortran -> C++ port, which I refer to on a regular basis.

